PR #7, opened 2019-08-07
Various fixes

alfaromeotango:
Add check for unsupported second engine
Return N/A when charging time is not applicable (not charging)
Last update icon does not exist. Changed to mdi-update
Binary Sensor naming collision for trunk and doors (locks vs doors)


arjenvrh:
Thanks for the fixes. It seems your work overlaps some fixes I checked in, so I took the liberty of merging the remaining fixes manually. 

Regards
Arjen

PR #17, opened 2020-01-12
Facilitating for Climatisation

andersop91:
Confirmed working sensor. I've also tried activating climatisation without luck. Disabling it works though.


andersop91:
More work done. Confirmed working climatisation, with 'electric' heating set as source!

arjenvrh:
Cool, thanks for the contribution!

PR #89, opened 2021-10-26
Update audi_services.py

albatorsk:
Updated configuration_endpoint URL for non-US. Credit goes to @MR-Staffman


pilot1981:
@mackay2k4 can also you look at issue #94 ?

PR #102, opened 2021-12-16
Adapting to extra_state_attributes instead of device_state_attributes

titiviking:
many log entries are shows for the different attributes and entities populated due to the deprication of device_state_attributes.


titiviking:
Linked to issue: https://github.com/arjenvrh/audi_connect_ha/issues/101

Issue #93, opened 2021-11-24
Set charge timer mode


zcopter:
Is there the possibility that this action be added on a future update? For example trigger the car to be set in timer mode when arriving at a specific location rather than manually having to set it on the app or mmi. I see it can be done from "shortcuts" but this would be a nice feature to automate the car to use timer mode without manually setting it every day.

t0bias-r:
@mpredfearn

These are the payloads required for Audi e-tron to program the charging timer (up to 5 timers) and enable/disable timer based charging.

**Timer programming**
POST to https://mal-3a.prd.eu.dp.vwg-connect.com/api/bs/profiletimer/v1/vehicles/WAUZZZ.../actions
Payload: `{ "action": { "type": "setTimers", "timerList": { "timer": [ { "timerID": 1, "activationState": "activated", "timerSettings": { "timerFrequency": "cyclic", "timerTime": "05:30", "weekdayMask": "yyyyyyy" }, "chargerSettings": { "chargeOption": "enabled" }, "climatisationOption": "enabled" } ] } }}`

**Enable timer based charging**
POST to https://fal-3a.prd.eu.dp.vwg-connect.com/fs-car/bs/batterycharge/v1/Audi/DE/vehicles/WAUZZZ.../charger/actions
Payload: `{ "action": { "type": "selectChargingMode", "settings": { "chargeModeSelection": { "value": "timerBasedCharging" } } }}`

**Immediate charing / disable timer based charging**
POST to https://fal-3a.prd.eu.dp.vwg-connect.com/fs-car/bs/batterycharge/v1/Audi/DE/vehicles/WAUZZZ.../charger/actions
Payload: `{ "action": { "type": "selectChargingMode", "settings": { "chargeModeSelection": { "value": "immediateCharging" } } }}`

mpredfearn:
@t0bias-r awesome, thank you! The current code forms an XML payload https://github.com/arjenvrh/audi_connect_ha/blob/master/custom_components/audiconnect/audi_services.py#L460 - should I chnange it to JSON, or do you suppose I can re-encode the payload in XML? I'll give it a try. Thanks again!

t0bias-r:
What I have seen in the Android apps (v4.5/v4.6) is that all actions are json encoded.
I guess that Audi switched the API from XML to Json and they turn off the deprecated XML support step by step. Like they did it with the login procedure.
So I would change to Json but I cannot garantuee that it works for all users.

[...]

PR #104, opened 2022-01-08
New items + fixes

t0bias-r:
- Added: shortterm_current, shortterm_reset, longterm_current, longterm_reset
- Added: outdoor_temperature
- Added: charging_mode
- Added: energy_flow
- Fixed: X-Client-ID in __get_headers()
- Fixed: parking position fix (error 204)
- Fixed: calculation of actualChargeRate

Tested with Audi e-tron MY2021


fhloston:
Before applying this PR my audi connect integration is broken, with this PR it works again. Thank you @t0bias-r !
```
This error originated from a custom integration.

Logger: homeassistant.config_entries
Source: custom_components/audiconnect/audi_services.py:241
Integration: Audi Connect
First occurred: 09:54:33 (1 occurrences)
Last logged: 09:54:33

Error setting up entry <redacted> for audiconnect
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 313, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/config/custom_components/audiconnect/__init__.py", line 106, in async_setup_entry
    return await data.update(utcnow())
  File "/config/custom_components/audiconnect/audi_account.py", line 162, in update
    if not await self.connection.update(None):
  File "/config/custom_components/audiconnect/audi_connect_account.py", line 113, in update
    vehicles_response = await self._audi_service.get_vehicle_information()
  File "/config/custom_components/audiconnect/audi_services.py", line 241, in get_vehicle_information
    "Authorization": "Bearer " + self.audiToken["access_token"],
KeyError: 'access_token'
```

custom_components/audiconnect/audi_connect_account.py
```diff
@@ -589,15 +600,22 @@ async def update_vehicle_charger(self):
                 self._vehicle.state["chargingState"] = get_attr(
                     result, "charger.status.chargingStatusData.chargingState.content"
                 )
-                self._vehicle.state["actualChargeRate"] = get_attr(
+                self._vehicle.state["actualChargeRate"] = float(get_attr(
```
dfigus, line 603: This fails for my car. In case there is no charge rate available it will be `None `and the float conversion will fail. Error will be `float() argument must be a string or a number, not 'NoneType'`.

Not sure whether my car supports it at all (A6 PHEV), but at least when it's not charging it is `None`. Furthermore down below you already do a parse float: https://github.com/arjenvrh/audi_connect_ha/blob/db71c2243b3713fe1edb7e1a6ae9cac9e74ec975/custom_components/audiconnect/audi_connect_account.py#L1094
Note sure why it's needed here again (just with the /10).

If I have time later I try to find a public charging spot to see whether this works during charging.

dfigus: Just checked when charging and my PHEV does not support this data, hence you need to guard this against `None`.

ddimick:
Eagerly anticipating the this merge!

mpredfearn:
@t0bias-r this looks great! As you appear to have the means to dissect the API calls made by the Audi app, please could I put in a request for https://github.com/arjenvrh/audi_connect_ha/issues/93 ? It would be really helpful to allow setting timer charge mode automatically when the car returns home rather than having to press the button on the MMI every time (and educating the significant other to do the same!)

custom_components/audiconnect/audi_connect_account.py
t0bias-r: I added a fix for this

t0bias-r:
> @t0bias-r this looks great! As you appear to have the means to dissect the API calls made by the Audi app, please could I put in a request for #93 ? It would be really helpful to allow setting timer charge mode automatically when the car returns home rather than having to press the button on the MMI every time (and educating the significant other to do the same!)

I could provide the json that has to be sent to server. Would that be helpful?

mpredfearn:
@t0bias-r yeah that would be awesome - I'm happy to add the code and do the PR if you could post the JSON.

derhappy:
@t0bias-r can you explain what the "reset" sensors are supposed to show? 
For our car (and the one we had before) the long term reset always throws an error because there is only one long term result. 
Short term reset is just the second latest entry from the short term trip data?

t0bias-r:
@derhappy `_current` is the current value displayed in the vehicle whereas `_reset` is the value that the on-board computer had when it was reset. ShortTerm (in my car) is automatically reset after 2 hours of inactivity, where as LongTerm must be reset manually.
If you never reset LongTerm manually, there was an error. It is fixed in pending PR188: https://github.com/arjenvrh/audi_connect_ha/blob/45beaeaac9189c7c8f6f85705a54c5f0fbb9c258/custom_components/audiconnect/audi_services.py#L313

PR #188, opened 2023-06-01
Fixed reading charge properties eq 0 and removed unnecessary error messages

t0bias-r:
- Fix: maxChargeCurrent eq 0, actualChargeRate eq 0, chargingPower eq 0
- Removed error message when a feature is detected as unavailable

PR #199, opened 2023-08-28
remove unexpected argument from device info

Gfeuillen:
This solves #197 for me


florie1706:
This change is already in PR #188

florie1706 (changes requested):
Already in PR #188

PR #203, opened 2023-10-03
Update audi_services.py

reisfni:
Fixes issue in Home Assistant where device/entities were not populating after adding the integration.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -250,7 +250,7 @@ class AudiService:
         }
         req_rsp, rep_rsptxt = await self._api.request(
             "POST",
-            "https://app-api.live-my.audi.com/vgql/v1/graphql",
+            "https://app-api.my.aoa.audi.com/vgql/v1/graphql",
             json.dumps(req_data),
             headers=headers,
             allow_redirects=False,


damienheiser (approved):
I've made the change to my HA and can confirm it works as desired.

florie1706:
I am in Germany. Old Url works, I changed it to the Url of this pull request anyway, but that did not work for me. So I am on the old Url again which still works for me.

agentadam-code:
WOW this works. Thanks so much.

oleg-d:
> I am in Germany. Old Url still worked, I changed it to the Url of this pull request anyway, but that did not work for me. So I am on the old Url again which still works for me.

Same for me, old URL was working and new one is not. I'm in Finland.

woopstar:
> > I am in Germany. Old Url still worked, I changed it to the Url of this pull request anyway, but that did not work for me. So I am on the old Url again which still works for me.
> 
> Same for me, old URL was working and new one is not. I'm in Finland.

Same for me in Denmark. Old url works, not new one.

victorkp:
US user here -- this updated URL works.

We should be switching URLs dependent the user's specified region.

victorkp:
I've made pull request #211, which only sets the new URL for US users. Non-US users will continue to use the old URL. @reisfni, you should close this PR.

PR #211, opened 2023-11-25
fix: Integration fails to setup

victorkp:
Fixes URL issues for US region users, while still retaining the old URL for non-US users.

This is a replacement for #203, which breaks the URL for non-US users.


derhappy:
AFAIK that’s how it was implemented some time ago. That’s why you have to give the region when you install the integration.

Issue #217, opened 2024-01-16
SOURCE_TYPE_GPS


over40:
Logger: homeassistant.components.device_tracker
Source: helpers/deprecation.py:205
Integration: Geräte-Tracker (documentation, issues)
First occurred: 15. Januar 2024 um 23:21:56 (2 occurrences)
Last logged: 15. Januar 2024 um 23:21:56

SOURCE_TYPE_GPS was used from audiconnect, this is a deprecated constant which will be removed in HA Core 2025.1. Use SourceType.GPS instead, please create a bug report at https://github.com/arjenvrh/audi_connect_ha/issues

Prash1407:
+1 additional info on how to solve:

This log entry from Home Assistant indicates that the audiconnect integration is using a deprecated constant, SOURCE_TYPE_GPS. This constant is scheduled to be removed in the Home Assistant Core 2025.1 release. The recommended action is to use SourceType.GPS instead.

its not urgent but possible a quick fix and can be released with a new update @arjenvrh

PR #223, opened 2024-02-01
Update device_tracker.py

Prash1407:
Recent log entry from Home Assistant indicates that the audiconnect integration is using a deprecated constant, SOURCE_TYPE_GPS. This constant is scheduled to be removed in the Home Assistant Core 2025.1 release. The recommended action is to use SourceType.GPS instead.

See also #217 #222 #221


Prash1407:
> I actually just made this change. It is into the master now.

perfect, saw and tested - all works fine here in CH

PR #229, opened 2024-03-02
partial update to new Audi API

cdnninja:
t0bias-r initial work to fix the integration.


cdnninja:
@t0bias-r  thoughts on merging into this repo?    If you plan to invest in this longer term I am willing to support where I can.  Could see if @arjenvrh is willing to give us admin and convert this to a "org" which makes it more a shared project for longer term.

PR #230, opened 2024-03-02
DivingUp Connected to the new Api... more robust handling of electric and non-electric vehicles

DivingUp:


DivingUp:
Update to the new new API... seems to work for electric and non-electric cars... Data is read out, unlocking and locking the car works. Parking heater works. I could not test the rest.

DivingUp:
If you remove the comment in audi_services.py line 189, you can directly see all the data that the Audi api spits out for the vehicle in the HA log

cdnninja:
Let me know if you want me to merge.  Saw some feedback in the thread.

PR #253, opened 2024-03-05
fix: charging time feat: Added new sensors

Ingrimmsch93:
fixed:
remaining charging time to complete

added:
externalPower => Status of connected external power source
plugLockState => Status of locking function of charging cable
plugledColor => Color of status LED next to charging port


Kolbi:
Again my question: Why and where do we need the dashboard.py?

Ingrimmsch93:
I don't get your point.. What do you mean?

dashboard.py defines the entities in Home Assistant.
It is located in audiconnect folder of your component installation.

Kolbi:
I got it thanks, I'm still sometimes confused about this HACs package as it is not splitted in backend and frontend.

PR #255, opened 2024-03-06
Deactivate Short/longTrip to prevent 403errors

DivingUp:


cdnninja:
@t0bias-r  You mentioned another solution to this.   I am not familiar yet with the code to do that quickly What would turn around time be for you?   If will be a bit I may merge this to not have accounts blocked in the meantime.

Kolbi:
+1 - I would deactivate it until fix.

PR #265, opened 2024-03-07
fix: tidy up code + update readme

Kolbi:


t0bias-r:
@Kolbi  looks good to me 👍

PR #277, opened 2024-03-10
chore: inactive issues

cdnninja:
Creating a bot that comments on old issues and marks as stale.  I plan to set this to close issues after 90 days of inactivity.  However wanting to start with it just commenting on them.  Make sure that works and set the -1 to 90.


Kolbi (approved):
Makes sense for me. 

We just shouldn't forget about changing the days-before-issue-close later on :)

PR #278, opened 2024-03-10
chore: Lint Checks in CI

cdnninja:


Kolbi:
Isn't it doing mostly the same like pre-commit does?

cdnninja:
Yes should match.   Not all pull requests come from visual studio. This shows if it conforms to the pre commits or not.

cdnninja:
I am seeing now the pre-commit shows in the checks. I will need to read more as I didn't see that happening previously.  Maybe we are trying to do the same thing just two different ways.

Issue #186, opened 2023-05-15
Multiple accounts


retrov:
Been trying to add a second account to the integration but it seems to crash due to duplicate sensor names.
So when I go to the new added device it actually goes to the first car that was added.
Anyone seen this or have a way around it ?

Also tried to rename the entities before adding the second..

retrov:
Yes.

cdnninja:
To clarify you have two accounts with different cars?

retrov:
Two accounts, one for each car. But since I'm a user of both I have access to both cars from one login. So it starts mixing up the entities from both cars. One entity might be from one car and another one from the other.

retrov:
Tried it again today, and it's still picking up entities from each car and presenting them as one. Expected behaviour would be two devices with each of its entitites. It seems to have picked up all of the entities of my Etron 50, but the entities that this car doesn't have, ie Sunroof, is picked from my Etron 55.

retrov:
I've attached the debug from the init of the plugin, I see that it get's the data from both cars.

[audiconnect_debug.txt](https://github.com/audiconnect/audi_connect_ha/files/14817533/audiconnect_debug.txt)

retrov:
Anyone else seeing this with multiple cars on the account?

retrov:
Yes, still a problem for me too.  It was working for a short time last year but haven't been able to replicate it with the older versions.

PR #287, opened 2024-03-12
fix: Services.yaml // New Identifier

Kolbi:
Currently need some testing as I'm changing the identifier: 
Instead of:
"identifiers": {(DOMAIN, self._instrument.vehicle_name)},
We now use:
 "identifiers": {(DOMAIN, self._instrument.vehicle_vin)},

This makes the Services.yaml working smooth and a name can't be a unique id.

But side-effects has to be tested. I had two vehicles after the change but might be my settings related.

I thought maybe if we update the identifier but I couldn't find an easy solution for it.

### DIFF
--- a/custom_components/audiconnect/__init__.py
+++ b/custom_components/audiconnect/__init__.py
@@ -4,6 +4,9 @@ from datetime import timedelta
 import voluptuous as vol
 
 import homeassistant.helpers.config_validation as cv
+from homeassistant.helpers import device_registry as dr
+from homeassistant.config_entries import ConfigEntry
+from homeassistant.core import HomeAssistant
 from homeassistant.util.dt import utcnow
 from homeassistant import config_entries
 from homeassistant.const import (
@@ -122,4 +125,11 @@ async def async_unload_entry(hass, config_entry):
 
     del hass.data[DOMAIN][account]
 
+    return True
+
+
+async def async_remove_config_entry_device(
+    hass: HomeAssistant, config_entry: ConfigEntry, device_entry: dr.DeviceEntry
+) -> bool:
+    """Remove a config entry from a device."""
     return True
--- a/custom_components/audiconnect/audi_account.py
+++ b/custom_components/audiconnect/audi_account.py
@@ -2,7 +2,7 @@ import logging
 from datetime import timedelta
 import voluptuous as vol
 
-import homeassistant.helpers.config_validation as cv
+from homeassistant.helpers import config_validation as cv, device_registry as dr
 from homeassistant.helpers.dispatcher import (
     async_dispatcher_send,
 )
@@ -32,7 +32,8 @@ from .const import (
 
 REFRESH_VEHICLE_DATA_FAILED_EVENT = "refresh_failed"
 REFRESH_VEHICLE_DATA_COMPLETED_EVENT = "refresh_completed"
-SERVICE_REFRESH_VEHICLE_DATA = "refresh_data"
+
+SERVICE_REFRESH_VEHICLE_DATA = "refresh_vehicle_data"
 SERVICE_REFRESH_VEHICLE_DATA_SCHEMA = vol.Schema(
     {
         vol.Required(CONF_VIN): cv.string,
@@ -168,7 +169,9 @@ class AudiAccount(AudiConnectObserver):
             )
 
     async def execute_vehicle_action(self, service):
-        vin = service.data.get(CONF_VIN).lower()
+        device_id = service.data.get(CONF_VIN).lower()
+        device = dr.async_get(self.hass).async_get(device_id)
+        vin = dict(device.identifiers).get(DOMAIN)
         action = service.data.get(CONF_ACTION).lower()
 
         if action == "lock":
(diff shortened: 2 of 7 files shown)


cdnninja:
I will test this out when I get a chance.  We may be able to also do a config migration on this. https://developers.home-assistant.io/docs/config_entries_config_flow_handler/#config-entry-migration

cdnninja:
I haven't had a chance to look at this yet been a busy week.

coreywillwhat:
I reviewed, tested, and merged into my fork. Absolutely wonderful having the Device drop down in the service calls. Thanks for doing this!

Kolbi:
> I reviewed, tested, and merged into my fork. Absolutely wonderful having the Device drop down in the service calls. Thanks for doing this!
> 
> Note: As you mentioned, i also had a blank, duplicate device, where the "Room" was still attached. Seems its creating a new device and removing the details from the previous, other than the room. The new device did not have a room assigned.
> 
> I deleted the integration > restarted HA > readded the AudiConnect integration, and then only had 1 vehicle.
> 
> Otherwise this worked like a charm.

I was trying to write an migration definition: https://github.com/Kolbi/audi_connect_ha/blob/b075dbd16cce736be4272f9487eeda99b9b27012/custom_components/audiconnect/__init__.py

but it doesn't work.... I even don't see in any entries from it in the logs....

cdnninja:
I hope to test this soon.  Sorry crazy busy with the toddler and baby!   

For migration here is what I did on another project when had the same issue. https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/894299ab58774c08f3a6f7b66fa24229f06799ef/custom_components/kia_uvo/__init__.py#L75-L123

Probably could be cleaner but it worked. 

I assume this would be a breaking change? If so we can rename it so the release engine will make this a 2.x release and label is breaking.

Kolbi:
I know this PR is not needed anymore but want to let it open until we finished the "async_migrate_entry" topic and implemented it in @coreywillwhat PR.

I still have no clue why it is not working, but currently i don't have any testing enviroment which makes testing a little bit more difficult. :)

coreywillwhat:
I tried your code block @Kolbi . Having the same issue where it never initiates the change/no logs. 

A little more digging shows the version is never set for the entry in the config flow upon initial integration setup, so when you're looking for `if config_entry.version == 1:` maybe it isn't meeting the criteria?

In the [AudiConfigFlow](https://github.com/audiconnect/audi_connect_ha/blob/master/custom_components/audiconnect/config_flow.py#L31) I think we need to define the version similar to how @cdnninja did in his [referenced hyundai/kia integration](https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/894299ab58774c08f3a6f7b66fa24229f06799ef/custom_components/kia_uvo/config_flow.py#L134-L137).

A good starting place might be to figure out what version it's been assigning, if anything, then account for that, and assign the new `VERSION = 2` in the config_flow.py so we dont run into this in the future.

Kolbi:
I'm stuck at the point with migration what I don't know how to access the already (?) saved VIN from the vehicle....

Kolbi:
> Edit: FWIW -- This time around I tried deleting the duplicate device and I haven't encountered any issues. the "new" device maintains the correct `entity id`'s so existing references aren't impacted.
> 
So worst case no migrate but explaination how to delete the old device.

coreywillwhat:
> I'm stuck at the point with migration what I don't know how to access the already (?) saved VIN from the vehicle....

Could you use a [listener](https://developers.home-assistant.io/docs/integration_listen_events/) to wait for Home Assistant to finish starting (EVENT_HOMEASSISTANT_STARTED) and then callback to finish the migration using the vin from AudiConnectAccount?

cdnninja:
What I did in kia_uvo is the config migration just deleted the entities.   That also removed historical data though.   It did mean if nothing tied to it everything come up clean.  Not perfect but may work. Would make this PR a breaking change so it would be 2.x.

Kolbi:
@cdnninja If you want to test use repo: https://github.com/Kolbi/audi_connect_ha/ and install version: https://github.com/Kolbi/audi_connect_ha/releases/tag/v2.0.0a1

yeah maybe breaking-change but maybe even without automatic migration or deleting of anything as the data remains only a second empty device appears.

cdnninja:
Yes I plan to test.

Kolbi:
https://github.com/audiconnect/audi_connect_ha/issues/331 as mentioned here maybe we should change the "unique_id"s

cdnninja:
Reviewing this in more depth and testing it out I think two things should happen: 

1. Unique ID should be fully unique, including both VIN and Username.   This covers car and account: https://github.com/audiconnect/audi_connect_ha/issues/186
2. We will need to get migration sorted out as it should clean up behind the changes to atleast ensure no duplicate devices.   

Let me know thoughts?

Kolbi:
Do we have limitation of length for the unique id? Need to check the documentation. 
About the migration maybe someone else can take over that part?

cdnninja:
Sounds good.  I'll see if I can get this working.   I don't have a ton of time but will give it a go.  May take me a bit.

PR #293, opened 2024-03-15
fix: update *.bff.cariad.digital endpoints for different regions

SamKirsch10:
At least for me in the US, I have to use the `na` endpoint. Found this by proxying my app and seeing this instead. Previously I was getting 401 unauthorized in my logs and essentially no data back.

Confirmed I'm now getting all expected entities:
image


cengel74:
I'm in the US as well and this fixed my issue.  Was only getting lock status before I applied this change, now I see all the expected sensors

cdnninja:
Thank you for this! Great to see others chipping in to solve issues.   I tested this in Canada and works here.

cdnninja (approved):

Issue #299, opened 2024-03-20
Last update is a string instead of a datetime.


wigster:
The sensor.<name>_last_update state is interpreted as a string instead of a datetime. 

Possibly, it is just missing the 'device_class: timestamp' attribute. 

As it is, the sensor does not get reformatted to e.g. relative time by the UI.

Kolbi:
Can you check https://github.com/audiconnect/audi_connect_ha/pull/300 ?

wigster:
It doesn't seem to solve the problem. In the entities card:

![image](https://github.com/audiconnect/audi_connect_ha/assets/10956987/0a3b850c-190d-4684-a406-1a913d363174)

I still do not have the device_class in the attributes of the entity. I have checked that the dashboard.py contains your latest changes from #300

wigster:
Yep can confirm it works in your branch

Kolbi:
I close this case, as I guess the https://github.com/audiconnect/audi_connect_ha/pull/300 will be merged today or tomorrow @cdnninja

PR #300, opened 2024-03-20
fix: LastUpdate timestamp Bug 299

Kolbi:
#299

### DIFF
--- a/custom_components/audiconnect/manifest.json
+++ b/custom_components/audiconnect/manifest.json
@@ -9,5 +9,5 @@
   "issue_tracker": "https://github.com/arjenvrh/audi_connect_ha/issues",
   "loggers": ["audiconnect"],
   "requirements": ["beautifulsoup4"],
-  "version": "1.0"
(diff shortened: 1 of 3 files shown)


custom_components/audiconnect/manifest.json
```diff
@@ -9,5 +9,5 @@
   "issue_tracker": "https://github.com/arjenvrh/audi_connect_ha/issues",
   "loggers": ["audiconnect"],
   "requirements": ["beautifulsoup4"],
-  "version": "1.0.0"
+  "version": "1.0.1"
```
cdnninja, line 12: The release CI will do this.  This will cause it to skip a number or maybe even not generate a release.

cdnninja (changes requested):

custom_components/audiconnect/manifest.json
Kolbi: It was done automatically on my github :D but removed the change manually

PR #305, opened 2024-03-24
fix: Update Refresh Interval

coreywillwhat:
Excessive polling will lock users out of their accounts both in HA and in the Audi App. Minimum should be 15 minutes.


cdnninja:
Thank you for this!

Kolbi:
Am I right that already saved accounts remain with the existing SCAN_INTERVAL? So we need to manually adapt it to 15minutes or readd the audi account?

coreywillwhat:
That's correct: Manually adapt or remove/readd integration. We'll work on a "Configure" option for the integration to update the scan interval on the fly.

Kolbi:
> > Am I right that already saved accounts remain with the existing SCAN_INTERVAL? So we need to manually adapt it to 15minutes or readd the audi account?
> 
> In which file is this information? I don't wanna readd everything. Thank you!

.storage/core.config_entries

@coreywillwhat maybe if we get "async def async_migrate_entry": https://github.com/audiconnect/audi_connect_ha/pull/287#issuecomment-2022776849 working we can also set this variable to default 15.

coreywillwhat:
@Kolbi @moritzmaier84 with #312 I've also added an OptionsFlowHandler. Options are stored separately than the original data entry. Therefore I wrote a simple hierarchy to take the options value first, then the originally entered data, then default. So if the user updates the scan interval from the front end, there will be no need to migrate the data.

In short: If a user updates to this PR #312 , they can then go to the integration, click configure, and change the scan interval whenever they want.

PR #325, opened 2024-04-03
feat: New Service - Refresh Cloud Data

coreywillwhat:
### `audiconnect.refresh_cloud_data`

Introduces the `audiconnect.refresh_cloud_data` service call. The service leverages an existing backend function, now made accessible through the Home Assistant UI. It's designed to update vehicle data from the cloud for all vehicles linked to an account **without** initiating a vehicle-side refresh.

### Use Case:
Given some limitations with the "Refresh Vehicle Data" functionality, this service call provides a workaround by allowing users to manually trigger data updates. It's particularly useful for users who prefer to extend the default refresh interval significantly and wish to manually request updates at specific moments, such as after starting climate control or during periods of inactivity.

### Service Call Description:

- Functionality: Updates data for all vehicles from the cloud, mirroring the action performed at integration startup or during scheduled refresh intervals.
- Behavior: Does not force a vehicle-side data refresh. Consequently, if vehicles haven't recently pushed updates, retrieved data might be outdated.
- Recommended Usage: Ideal for post-command updates (e.g., after initiating climate control). To ensure data accuracy, a delay of approximately 30 seconds is recommended between command issuance and this service call.
- Note: This service essentially replicates the function of restarting the integration or polling for updates, offering a more granular control over data refresh moments.

This addition aims to enhance user control over data updates, addressing specific scenarios where updated information is crucial without unnecessarily frequent automated refreshes.

### DIFF
--- a/custom_components/audiconnect/audi_account.py
+++ b/custom_components/audiconnect/audi_account.py
@@ -63,6 +63,8 @@ SERVICE_START_CLIMATE_CONTROL_SCHEMA = vol.Schema(
     }
 )
 
+SERVICE_REFRESH_CLOUD_DATA = "refresh_cloud_data"
+
 _LOGGER = logging.getLogger(__name__)
 
 
@@ -102,6 +104,11 @@ class AudiAccount(AudiConnectObserver):
             SERVICE_START_CLIMATE_CONTROL,
             self.start_climate_control,
             schema=SERVICE_START_CLIMATE_CONTROL_SCHEMA,
+        )
+        self.hass.services.async_register(
+            DOMAIN,
+            SERVICE_REFRESH_CLOUD_DATA,
+            self.update,
         )
 
         self.connection.add_observer(self)
--- a/custom_components/audiconnect/strings.json
+++ b/custom_components/audiconnect/strings.json
@@ -117,6 +117,10 @@
           "description": "(Optional) Enable or disable Auto Seat Comfort for the rear-right seat."
         }
       }
+    },
+    "refresh_cloud_data": {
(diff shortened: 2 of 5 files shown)


cdnninja:
This looks to be missing the [services.yaml](https://github.com/audiconnect/audi_connect_ha/blob/master/custom_components/audiconnect/services.yaml) config.

coreywillwhat:
> This looks to be missing the [services.yaml](https://github.com/audiconnect/audi_connect_ha/blob/master/custom_components/audiconnect/services.yaml) config. 

Since there are no user inputs it only uses the strings/translations. No services.yaml needed

cdnninja:
Maybe too much future proofing but if no selection on vehicle what happens if integration is setup twice for two accounts? One user has that today I believe.

coreywillwhat:
hmm interesting I hadn't thought about 2 different accounts/configurations. I have only assumed multiple vehicles under the same account. In that instance an account selection would probably make sense? This is the same as integration start/polling so all vehicles under the same account would be updated. I'm not sure how it would work with 2 accounts.

Open to more of your thoughts on this.

Kolbi:
I guess as DOMAIN should be for both accounts "audiconnect" all accounts & VINS would get updated, but I think that would be okay.

coreywillwhat:
That would be my expectation as well, but if that's the case, I should update the documentation to state that all "_accounts_" and vehicles will be updated.

Kolbi:
I just merged the PR - all vehicle for me automatically means all accounts as well, for me ;)

PR #321, opened 2024-04-02
feat: Refresh Interval Options

coreywillwhat:
## Overview

This update introduces several enhancements and fixes aimed at improving the integration's configuration flexibility and efficiency in polling mechanisms.

### Enhancements
- **Configurable Scan Interval**: Integration of `OptionsFlowHandler` allows users to adjust the scan interval via the `Configure` menu, enhancing user control.
- **Efficient Polling**: Adoption of `async_track_time_interval` for polling initiation at startup removes the need for manual scheduling in `AudiAccount.update`, streamlining the update process.
- **Localization Updates**: Made updates to strings and translations to ensure clarity and accessibility.

### Technical Notes
- **Restart Requirement**: Changes necessitate a system restart to take effect. This requirement is detailed in the `data_description` within translations, ensuring users are informed via the UI.
- **Polling Mechanism Overhaul**: Leveraging `async_track_time_interval` from the get-go allows for interval-based polling using Home Assistant's built-in scheduling tools, bypassing the need for hardcoded scheduling.
- **Future-Proofing `AudiAccount.update`**: With these changes, `AudiAccount.update` is now freed up for potential use in front-end service calls, with plans for a follow-up PR pending this update's approval and merge.
- **Effort Towards User-Controlled Polling**: Next steps involve utilizing the `coordinator` for more flexible update scheduling, aiming to give users the ability to toggle polling on and off directly from the UI.

PR #327, opened 2024-04-03
fix: Request Vehicle Data

coreywillwhat:
This update doesn't directly fix the `request_vehicle_data` service issue but improves the logic flow:
- **Current Logic**: A successful refresh triggers a cloud update. A failed refresh ends the process.
- **Updated Logic**: Now, even if the refresh fails, the system will give the vehicle an opportunity to "check-in" with the cloud before polling for an update. 
 
This approach has shown effectiveness in my testing for obtaining updated vehicle/cloud data following an action request.


cdnninja:
Wouldn't this tie up a HA polling thread and also duplicate the built in HA polling engines logic that does this already?

coreywillwhat:
In my testing I found that executing vehicle commands, update_cloud_data and active polling simultaneously doesn't cause any interference. All requests for update succeed. Maybe if we were using a coordinator that would be an issue? I'm not sure.

Also, it's worth noting that the refresh_vehicle_data service is already using self.update for the update if it succeeds. A method that's been in use since before I began working on this. This means the functions in question have been coexisting without issue for some time. Let me know if I'm misunderstanding your concerns though.

This primarily solves the issue that users aren't currently getting updates from their vehicle after their commands to start climate control, or lock/unlock etc. The commands go through, but since the request to wake the vehicle isn't working, the data doesn't update until the next interval data poll, which could be a while.

Open to more thoughts though if you have further concerns.

coreywillwhat:
I'd also like to point out I'm very outside my element, so i'll defer to your judgement with any of these. You're the captain, i'm just swabbing the decks. haha

Kolbi:
Maybe have a look into: https://github.com/audiconnect/audi_connect_ha/discussions/282 I guess there you can see some examples how to retrieve action answers.

coreywillwhat:
I probably over explained and just caused confusion:
The "update" is how refresh vehicle data retrieves the updated data. This PR is just telling the system to update the cloud data even if the vehicle 'wake' function fails.

coreywillwhat:
At the risk of telling you something you already know...and explaining too much again haha: 
When `execute_vehicle_action` or `start_climate_control` is requested, the flow currently requests the vehicle action, once it gets a succeeded or failed message, the flow then requests the vehicle to 'wake' and check-in with the cloud. If the 'wake' function succeeds, the flow then requests a cloud update to get the latest vehicle data. If the 'wake' function fails, the process ends and no cloud update is requested.

Our 'wake' function, also known as `refresh_vehicle_data`, currently isn't working. And since the wake function is failing, the flow isn't polling the cloud for updated data and the user can't see that the climate control turned on or the doors were locked until the next active polling interval. 

Ultimately the 'wake' function really isn't needed as a vehicle will automatically 'check-in' with the cloud after being sent an action. Therefore, even if our wake function fails, we should request a cloud update.

This PR is just telling the system to update the cloud data regardless of the vehicle wake function succeeding or failing.

Kolbi:
I think I got it :)

cdnninja:
Isn't this the same function used for standard polling? I'm concerned multiple items were merged while discussion was still taking place on them.

coreywillwhat:
`self.update` is used for standard polling, which has not been modified, but is called here.
`refresh_vehicle_action` is not used for standard polling.

cdnninja:
So a bit less risk around worker jobs.   

Have you read this by any chance? https://developers.home-assistant.io/docs/asyncio_working_with_async/

cdnninja:
My concern is this change may lock up the event loop in home assistant until sleep interval ends. I'm no developer so not certain.  

If it locks event loop less busy home assistant may not notice but busy ones may.

coreywillwhat:
Gotcha, your concerns seems valid. Thanks for sharing that documentation. I read it through and seems like we would be okay, but as you said, i'm not a dev either so I can't be certain.

In my testing, I've run during scheduled intervals, the service call, and a vehicle action at the same time, and during the sleep pause and none have been interrupted or caused any issues. My system is fairly busy, but I can't say there isn't someone out there with more. Of note, the API/Login methods also use asyncio.sleep() function in the same manner for timeout.

To mitigate concerns, we could also remove the sleep interval entirely. I've reduced it to 10 seconds in my own instance and the data is appropriately updated even at the reduced interval.

Since the flow waits for the vehicle action succeeded or failed message, the sleep pause may not be required at all, as the vehicle has to check-in to say it succeeded or failed, so we could likely perform the cloud update immediately.

TL;DR
We may be better off trying this without the asyncio.sleep(UPDATE_SLEEP). Upside is we reduce latency of reporting the updated states, and reduce risks associated with asyncio locking up. Downside is possibility of miss an entity update, which is already the issue to begin with. So even if this doesn't work without delay, the end user won't know the difference.

I think this is the path forward but let me know what you think.

coreywillwhat:
Using my works instance of ChatGPT, here is what it said. Take it with a grain of salt of course:

"No, asyncio.sleep() does not risk locking up other functions when used in an async function within an asyncio-based application. This is one of the core advantages of using asyncio and writing asynchronous code. When you await asyncio.sleep(), it simply schedules a pause in the coroutine's execution, allowing the event loop to run other tasks that are ready to proceed.

Here’s a brief explanation of how it works:

    Non-Blocking: asyncio.sleep() is a non-blocking sleep. Unlike time.sleep(), which would block the entire thread and prevent any other code from running, asyncio.sleep() yields control back to the event loop. This means other coroutines can run during this wait.

    Cooperative Multitasking: Asyncio is based on cooperative multitasking, where tasks voluntarily yield control to run other tasks. Awaiting asyncio.sleep() is one way to yield control. This allows for efficient multitasking without the need for multiple threads or processes.

    Event Loop: The event loop is at the heart of asyncio applications. It keeps track of all the running tasks and executes them when they are ready. When a task awaits something (like asyncio.sleep()), the event loop schedules it to resume after the await completes and moves on to run other tasks in the meantime.

Therefore, using asyncio.sleep() in your async functions is the recommended way to introduce delays without affecting the responsiveness of your application. It allows your application to remain responsive and to continue performing other tasks, like handling user input, executing other coroutines, or responding to events, during the wait."

cdnninja:
@Kolbi no need to revert.  All good. Just wanted to share my thoughts. :)

So based on the above we may be okay.   The method is async so it should be fine.  Now I wonder the difference to the HA schedule method in the link vs applying a sleep.

coreywillwhat:
Its a good discussion, so i appreciate you challenging ideas.

Through all of this, i also noticed we unintentionally fixed another bug which likely caused excessive polling and lockouts for users at times:

In implementing a fix under PR #321 , we inadvertently resolved another issue. Previously, our configuration of `update` leveraged [async_track_point_in_utc_time](https://developers.home-assistant.io/docs/integration_listen_events/) to schedule subsequent updates post-each update, primarily for active polling.

**Identified Bug**: The `refresh_vehicle_data` service, which could be triggered by users or automatically after vehicle actions, called the `self.update` method. This resulted in excessive scheduling of `async_track_point_in_utc_time` callbacks - after each vehicle action, during polling intervals, and whenever the `refresh_vehicle_data` service was invoked. This over-scheduling could have overwhelmed the api in certain instances.

**Unintentional Fix**: While adjusting to `async_track_time_interval` in `async_setup_entry` for PR #321 , rather than within the update method, we unknowingly corrected this issue by eliminating the redundant scheduling of updates.

Issue #30, opened 2020-05-01
Support folding top (convertible)


wmalgadey:
Thank you very much for this integration. I did start my own, but didn't finish it to this level because of busy working days :(

Would it be possible to add support for the state of the folding top? I have only a german account, but it is available in webview and in the app:

![image](https://user-images.githubusercontent.com/141838/80793584-d3535d80-8b97-11ea-97a9-f29a471555a3.png)

arjenvrh:
If you can find out which of the properties contains this info and which values correspond to open and closed, I'd be happy to add it. As my top doesn't fold I can only guess. :-)

You could add a print statement in audi_connect_account.py in the function update_vehicle_statusreport() to get the full result of the call to self._audi_service.get_stored_vehicle_data. One of the fields is probably the one you want.

wmalgadey:
Thanks for the hint! I will look what I can do. The component looks complicated, tried to get a head of it, but have been busy since.

Do you know if somebody maybe did "refactor" the web ui? It seems to use a different API?

arjenvrh:
Closing this for now, when someone can provide me with the information I'd be happy to add it

Issue #333, opened 2024-04-05
Make mileage sensor total_increasing


wigster:
version: 1.1.0

The mileage sensor seems to be missing some attributes which would allow it to be included in long term statistics and then plotted using the statistics graph card. 

I believe it should have a state_class of total_increasing but maybe it’s also a question device_class?

Kolbi:
Hi @wigster makes totally sense for me, I already implemented it in https://github.com/audiconnect/audi_connect_ha/pull/334

Do you know other entities which would make sense for adding a state_class ?

wigster:
Im not sure it makes sense for the other entities I have — the range and tank level/battery level entities could be measurement state_classes, so that one could in principle use some SQL on the LTS data to process it, but I am not sure it would be that useful. 

I’ve made a bunch of template sensors working off those to work out gas mileage etc which is useful to store to look for any long term trends but that’s not really possible to calculate from historical tank levels.

PR #334, opened 2024-04-05
feat: state_class (#30)

Kolbi:
Introducte state_class for Sensor entities
- Possibility to define state_class for Sensors
- Currently only the Sensor mileage is defined with state_class="total_increasing",


Kolbi:
#333

custom_components/audiconnect/sensor.py
```diff
@@ -50,3 +50,8 @@ def device_class(self):
         ):
             return self._instrument.device_class
         return None
+
+    @property
+    def state_class(self):
+        """Return the state_class"""
+        return self._instrument.state_class
```
coreywillwhat, line 57: Should we validate the state_class similar to how `device_class` is done just above this by importing STATE_CLASSES from the same?
https://github.com/home-assistant/core/blob/dev/homeassistant/components/sensor/const.py

of note: looks like DEVICE_CLASSES is deprecated. I will put together a PR to fix that.

Kolbi: Good hint yes we should do it that way.

coreywillwhat:
I've tested your code in combination with the validator discussed and it works well. I can add the validator to this if you want, but dont want to overstep either.

It would also be good to add `device_class` for these so we can work on those where needed. I can do that in a separate PR, or you can add to this one.

Let me know.

Kolbi:
Please feel free to contribute. You may even edit my PR if you like. I plan being offline from Pc during weekend.

coreywillwhat (approved):
- added individual `device_class` for all `Sensors`
- added `state_class` validation

Kolbi:
Maybe device_class distance could be useful for some sensors. 

Need to check at Pc if others would also make sense:
https://developers.home-assistant.io/docs/core/entity/sensor/

But Prio low from my side currently.

coreywillwhat:
I agree it will be useful to update some of the sensors with device_class when we have time. Figure if we rock this in now we can easily go in and update values whenever later.

Thanks for this PR!

custom_components/audiconnect/dashboard.py
```diff
@@ -390,178 +393,302 @@ def create_instruments():
...
+            unit=None,
+            state_class=None,
+            device_class=None,
+        ),
+        Sensor(
+            attr="mileage",
+            name="Mileage",
+            icon="mdi:speedometer",
+            unit="km",
+            state_class="total_increasing",
```
cdnninja, line 409: I would think we should be importing these up front vs text and comparison. 

For example:

`from homeassistant.components.sensor import (
    SensorDeviceClass,
    SensorEntity,
    SensorEntityDescription,
    SensorStateClass,
)`

followed by: 

`state_class=SensorStateClass.TOTAL_INCREASING`


Examples are shown here: https://github.com/Hyundai-Kia-Connect/kia_uvo/blob/master/custom_components/kia_uvo/sensor.py

custom_components/audiconnect/dashboard.py
```diff
@@ -390,178 +393,302 @@ def create_instruments():
...
             icon="mdi:oil",
             unit="km",
+            state_class=None,
+            device_class=None,
+        ),
+        Sensor(
+            attr="oil_level",
+            name="Oil level",
+            icon="mdi:oil",
+            unit="%",
```
cdnninja, line 472: Unit should be imported from HA as well.

custom_components/audiconnect/dashboard.py
```diff
@@ -390,178 +393,302 @@ def create_instruments():
...
             unit="%",
+            state_class=None,
+            device_class=None,
         ),
         Sensor(
-            attr="charging_power", name="Charging power", icon="mdi:flash", unit="kW"
+            attr="charging_power",
+            name="Charging power",
+            icon="mdi:flash",
+            unit="kW",
```
cdnninja, line 568: Unit here should also be imported.

custom_components/audiconnect/dashboard.py
```diff
@@ -390,178 +393,302 @@ def create_instruments():
...
+            state_class=None,
+            device_class=None,
         ),
-        Sensor(attr="range", name="Range", icon="mdi:gas-station", unit="km"),
         Sensor(
             attr="hybrid_range",
             name="hybrid Range",
             icon="mdi:gas-station-outline",
             unit="km",
+            state_class=None,
```
cdnninja, line 433: Any reason to add None when it defaults to this?   Feels to me like longer code.

cdnninja (changes requested):

custom_components/audiconnect/dashboard.py
coreywillwhat: I added the defaults after the fact. which caused it to be redundant. I'll clean those up.

custom_components/audiconnect/dashboard.py
coreywillwhat: on it

custom_components/audiconnect/dashboard.py
coreywillwhat: oh yeah, this is much cleaner. I'll clean it up.

custom_components/audiconnect/dashboard.py
```diff
@@ -390,178 +409,227 @@ def create_instruments():
...
             attr="parking_light",
             name="Parking light",
             device_class="safety",
             icon="mdi:lightbulb",
         ),
-        BinarySensor(attr="any_window_open", name="Windows", device_class="window"),
+        BinarySensor(
+            attr="any_window_open",
+            name="Windows",
+            device_class="window",
```
cdnninja, line 631: This would be BinarySensorDeviceClass.WINDOW imported from 

`from homeassistant.components.binary_sensor import (
    BinarySensorDeviceClass,
    BinarySensorEntity,
    BinarySensorEntityDescription,
)`

custom_components/audiconnect/sensor.py
```diff
@@ -50,3 +50,13 @@ def device_class(self):
         ):
             return self._instrument.device_class
         return None
+
+    @property
+    def state_class(self):
+        """Return the state_class"""
+        if (
+            self._instrument.state_class is not None
```
cdnninja, line 58: I don't think this is needed anymore.

cdnninja (changes requested):

custom_components/audiconnect/dashboard.py
coreywillwhat: I stuck to just cleaning up the sensors @kolbi started with. Ran out of time and the entire file needs cleanup. Larger cleanup effort might be best under another PR.

Let me know

custom_components/audiconnect/sensor.py
coreywillwhat: Was torn on removing it or not since it's now a benign final check. I can remove it tomorrow

coreywillwhat: removed this and the device class check

custom_components/audiconnect/dashboard.py
coreywillwhat: binary_sensors are done

cdnninja (approved):

cdnninja:
Awesome thanks for doing this!!

Kolbi:
I'm not sure if this PR but with latest source and my patch branch in combination: https://github.com/Kolbi/audi_connect_ha/releases/tag/v2.0.0a7 I don't get any entities anymore....

coreywillwhat:
hmm thats odd, I've had them running for morning without issue. Around the time you commented this though, it started failing to login to the Audi Connect servers...

Kolbi:
Argh again API issues. Then it must have been bad timing then older version worked again some minutes later.

coreywillwhat:
Yeah idk, I rolled into another version and I log in now but still no data from the cloud. Idk if I'm just testing too much or what but the API has been awful this past week for me.

coreywillwhat:
> Argh again API issues. Then it must have been bad timing then older version worked again some minutes later. 

I set the audi_api timeout to 30 instead of 10 and it eventually pulled the data. Getting some serious latency of the data request... Thinking maybe we should build in an increase in the timeout with each retry.

cdnninja:
If latency why not just increase the timeout overall?  A retry plus longer timeout increases total time and increases call count.

coreywillwhat:
> If latency why not just increase the timeout overall?  A retry plus longer timeout increases total time and increases call count. 

Yeah I'm fried, I need a vacation haha was thinking timeout was a sleep. Different.

I think increasing timeout would make sense. This has happened a bunch recently, and I'm seeing similar latency in the Audi app at the same time. But the Audi app succeeds because longer timeout where HA fails.

Kolbi:
Updated to latest master - charging complete time entity works all other entities are deactivated. But can only continue on Monday sadly. But I moved to the old identifier with using the master so maybe that could cause the issue on my side currently.

Kolbi:
Seemed to be an issue on my side. Deleted whole integration and readded and all is working again. Still curious why in combination with my changes it was not working. But maybe because of the API issues.

Issue #318, opened 2024-04-01
Make remaining charge time a datetime?


wigster:
Would it make sense to make the remaining charge time an estimated datetime set for the latest predicted charge completion time as opposed to the raw number of minutes of charging remaining being reported by the API?

In this way, in the dashboard one could use it as a countdown as opposed to just updating every 10-15 minutes as the charging takes place and HA polls the integration.

coreywillwhat:
This is a good idea. I'll look into it.

coreywillwhat:
I've completed this and tested on my end and its working as expected. Would like input on functionality though.

Currently I'm using the `last_update` datetime as the anchor to calculate the estimated completion datetime, but this means once the `charge_time_remaining` hits 0, the new sensor would always display the time of the last update. Since the vehicle may update frequently or infrequently, once it is at 0 the displayed time wouldn't be very helpful. So i've also added that the sensor will return `None` or unknown if the remaining time is 0.

Let me know if you have a better idea.

wigster:
I guess it would be useful to be able to differentiate between having run to 0 remaining time and expecting to have finished, and not charging at all/not being known. The idea being that, with zero remaining, you can put “was supposed to have finished” as a message on the dashboard. 

So maybe a date time in the past seems ok to me. The sensor could be called something like “expected charge completion time” and if it’s in the past it is not unreasonable. 

For what it’s worth, My hybrid seems to update very often while it’s being charged, and very rarely when just parked.

coreywillwhat:
![image](https://github.com/audiconnect/audi_connect_ha/assets/104224685/df4dd0f2-a3f0-4d92-9791-8ad9b16f420d)
Name: **Charging Complete Time**
- I removed 'expected' from your suggested name as this wouldn't make sense if its in the past.
- Using "Charging" to align with the other sensors.
- I've updated the sensor to display the `datetime` regardless of in the past or not.
Thanks for the idea and input! I'll get a PR put together. There is one dependency PR ahead of this.

PR #335, opened 2024-04-05
feat: Sensor: Charging Complete Time

coreywillwhat:
**this PR is dependent upon PR #334 **

## New Sensor - Charging Complete Time
`sensor.charging_complete_time`
`device_class: timestamp`

This sensor displays the estimated completion timestamp for vehicle charging, as suggested in issue #318. This sensor aims to provide users with an accurate prediction of when their car's charging process will be completed, enhancing the overall user experience and providing valuable information for planning purposes.

### Implementation Details
- The sensor calculates the estimated charging completion timestamp based on the `last_update` timestamp . The `last_update` serves as an anchor point for this calculation.
- It utilizes the `remaining_charge_time` to estimate when the charging will be completed from the last known state update received from the vehicle.
- It's important to note that once the `remaining_charge_time` reaches 0, the sensor will display the timestamp of the last update received from the vehicle. If the vehicle checks-in later, while the charge time is still 0, the new `timestamp` will be the time it checked-in and no longer the time the vehicle finished charging.

### Considerations
- This approach ensures that the sensor remains functional and provides an estimate based on the latest available data, but it may not always reflect the exact moment the charging finished if updates from the vehicle are delayed or if the charging process concludes shortly after the last update.
- Users should be aware that the estimated completion timestamp is just that - an estimate. Delays in data transmission or changes in charging speed can affect the accuracy of this estimate.

### Rationale
This enhancement addresses a specific need/gap identified in issue #318, where the user expressed a desire for a calculated point in time, vs the existing `remaining_charge_time` which only displays an `int` of minutes. The gap in the previous sensor is that if the vehicle isn't updated, or "active polling" is turned off or has a large interval, the `int` minutes will not update regularly and for instance, could remain as 80 minutes even if the charging has already stopped.

This again, reduces the need for frequent active polling.


cdnninja:
Happy to review once merge conflicts are resolved but it should be good.

coreywillwhat:
> Happy to review once merge conflicts are resolved but it should be good.

This is all set!

PR #339, opened 2024-04-07
fix: Reduce duplicate debug logs

cdnninja:
I find the debug logs difficult to read due to the same data over and over ever scan.  This reduces that.


coreywillwhat:
Thank you! The logs have been brutal. This works great.

PR #356, opened 2024-04-10
feat: Enhanced Logging for _tryAppendStateWithTs and _tryAppendFieldWithTs

coreywillwhat:
This should help both devs and users being able to read the logs to understand when sensor values are found and appended, or not.
Q4 and e-tron owners are having issues seeing sensors, so this is the first piece to figuring that out.

### DIFF
--- a/custom_components/audiconnect/audi_models.py
+++ b/custom_components/audiconnect/audi_models.py
@@ -259,54 +259,47 @@ class VehicleDataResponse:
 
     def _tryAppendStateWithTs(self, json, name, tsoff, loc):
         _LOGGER.info(f"Looking for value for {name}.")
-        _LOGGER.debug(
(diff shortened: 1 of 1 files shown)


Kolbi:
Regaring "Also note that _LOGGER.info is reserved for the core, use _LOGGER.debug for anything else."
https://developers.home-assistant.io/docs/development_guidelines/?_highlight=log#log-messages

I changed LOGGER.info => LOGGER.debug

coreywillwhat:
Oh yeah great catch/find. Would be nice to use the .info level for higher level logging, but since we can't, I've removed some lines which don't feel useful anymore.

coreywillwhat:
removed f-strings from logging as well, as said in documentation.
ready for review.

Kolbi (approved):
looks good for me

Issue #352, opened 2024-04-10
Sun roof not available


over40:
Diese Entität wird nicht mehr von der Integration audiconnect bereitgestellt. Wenn die Entität nicht mehr verwendet wird, lösche sie in den Einstellungen.
![Screenshot 2024-04-10 114455](https://github.com/audiconnect/audi_connect_ha/assets/7571285/23c83730-27b8-4cd1-8d01-54a5bf35fccd)

Kolbi:
Can you provide debug log?

over40:
Hi Kolbi, i have a debug file here...
[2024-04-10 15_54_16.864 DEBUG (MainThread) [custom_components.audiconnect.audi_services] audiconnect - Car Data_ {_acces.log](https://github.com/audiconnect/audi_connect_ha/files/14933260/2024-04-10.15_54_16.864.DEBUG.MainThread.custom_components.audiconnect.audi_services.audiconnect.-.Car.Data_._acces.log)

Kolbi:
Thanks for the logfile.

Good thing:

>                {
                  "name":"sunRoof",
                  "status":[
                     "open"
                  ]
               }

Now need to check why it's not updated. :)

PR #358, opened 2024-04-11
fix: Patch sun roof

Kolbi:
Add entity sun_roof Open & Closed currently we only have closed in HA

fix https://github.com/audiconnect/audi_connect_ha/issues/352

### DIFF
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -899,7 +899,7 @@ class AudiConnectVehicle:
     @property
     def sun_roof(self):
         if self.sun_roof_supported:
-            return self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER")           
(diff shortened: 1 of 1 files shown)


coreywillwhat:
should be able to test existing vs this today.

cdnninja:
Oops missed that comment.   Let me know if I need to revert the merge.

coreywillwhat:
I'll let you know. I know from going through all the old issues, and closing them out, there is some code somewhere that accounts for the window/door not being available. So I wasn't sure if that was also something we needed to look into/account for as well.

coreywillwhat:
Sure enough, when the sunroof is open it reports as `unknown`
This patch fixes it. Thanks for this!

coreywillwhat:
spoke too soon. I think its always reporting `Open` now

coreywillwhat:
i've got it, i'll submit another PR

PR #360, opened 2024-04-11
fix: remove LOCK_STATE_HOOD

Kolbi:
It's not used.

And API doesn't have the value locked for the hood:
{
   "access":{
      "accessStatus":{
         "value":{
            "overallStatus":"unsafe",
            "carCapturedTimestamp":,
            "doors":[
               {
                  "name":"bonnet",
                  "status":[
                     **"closed"**
                  ]
               }


coreywillwhat:
You think there is a chance someone has this? i know the API reports different things for different cars, but in the same structure. Maybe we comment these out for now and see if anyone raises any flags? You have me hyper aware after breaking your any_window_open entity @Kolbi haha

Also, if you're confident, I'm fine to just remove it. Just curious your thoughts.

PR #367, opened 2024-04-11
fix: Treat SunRoof as Window

coreywillwhat:
### Summary
The `sunRoof` should be treated as a window.
#358 fixed the sunroof becoming unavailable, but during testing showed it was always returning `Open`. This fixes that and groups it with the other windows, following the same logic.

### Technical
- Moving `sun_roof` and `sun_roof_supported` to be grouped with the windows in `audi_connect_account.py`
- Using same logic as other windows (It is defined with the other windows in `audi_models.py > appendWindowState`)
- Updated `any_window_open` to also check the sunroof

_Tested and working as expected_

### DIFF
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -1074,7 +1074,13 @@ class AudiConnectVehicle:
         checkRightFront = self._vehicle.fields.get("STATE_RIGHT_FRONT_WINDOW")
         checkRightRear = self._vehicle.fields.get("STATE_RIGHT_REAR_WINDOW")
         checkSunRoof = self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER")
-        if checkLeftFront and checkLeftRear and checkRightFront and checkRightRear and checkSunRoof:
+        if (
+            checkLeftFront
+            and checkLeftRear
+            and checkRightFront
+            and checkRightRear
+            and checkSunRoof
+        ):
             return True
 
     @property
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -942,17 +942,6 @@ class AudiConnectVehicle:
             return True
 
     @property
-    def sun_roof(self):
-        if self.sun_roof_supported:
-            return self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER")
-
-    @property
-    def sun_roof_supported(self):
-        check = self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER")
-        if check is not None:
-            return True
-
-    @property
     def preheater_active(self):
         if self.preheater_active_supported:
             res = (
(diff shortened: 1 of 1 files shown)


Kolbi:
I still don't know why it should always be open for sun roof. 

value": "3" if status[0] == "closed" else "0",

Closed = 3
Open = 0
?

cdnninja:
The sensor is HA is looking for true false but also reverts to 0,1 being valid as well.   Anything outside or 0,1 doesn't map to true or false.  Not sure that helps, I haven't looked closely at this.

coreywillwhat:
`return self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER") != "3"`
So if sensor != closed (3) return True (Open)
If sensor = closed (3) return False

As far as I can tell, the gods forged this logic alongside fire and the wheel. I can't fight it or tell you why its designed this way.

coreywillwhat:
if you directly pass 3 to HA, wouldn't it be considered True (Open)?

cdnninja:
I had this pain in my Kia integration too.  Kia uses numbers that actually mean more than I thought surface level.   

A example that's made up 0 may mean closed 1 tilt 2 partial open 3 fully open. 

I had this with seats thinking it was just true false but only to find later some cars report the power level.

coreywillwhat:
I'm pretty sure you're right and the numbers were used in an old API and "3" was introduced around an issue with a 2 door car that didn't report rear doors/windows. The logic was probably adjusted at some poitn with an API change where, like now, the API is actually reporting "closed" and "open".

Might be worth a clean up at some point.
That said, this code is tested and working. Tried open, close, and tested other windows with the `any window open` logic and its all working as expected.

Given the back and forth I'll wait for either of you to approve/merge.

Kolbi:
Sorry I had no sun roof to test and have a lack of an testing environment. Sorry my home base is not Python :) I merged the PR.

coreywillwhat:
no worries at all, thanks @Kolbi

Kolbi:
> `return self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER") != "3"` So if sensor != closed (3) return True (Open) If sensor = closed (3) return False

Now i understand this point. :) I thought "return self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER")" returns the value from self._vehicle.fields.get("STATE_SUN_ROOF_MOTOR_COVER") but it doesn't :D it returns True or False :)

Kolbi:
My entity binary_sensor.audi_windows is not available anymore.

I guess because:
2024-04-12 08:27:17.252 DEBUG (MainThread) [custom_components.audiconnect.audi_models] Data that will be used for _tryAppendStateWithTs: [...] 'windows': [{'name': 'frontLeft', 'status': ['closed']}, {'name': 'frontRight', 'status': ['closed']}, {'name': 'rearLeft', 'status': ['closed']}, {'name': 'rearRight', 'status': ['closed']}, {'name': 'roofCover', 'status': ['unsupported']}, {'name': 'sunRoof', 'status': ['**unsupported**']}], [...]

PR #371, opened 2024-04-12
fix: Any Window Open w/ Unsupported Sunroof

coreywillwhat:
Allows for unsupported window state for the sunRoof.
I haven't had a chance to test this yet with my car. I will do that later today.
@Kolbi if you get a chance to test on yours that would be great.


Kolbi:
Seems not to work.

coreywillwhat:
ah okay, right, we can default to None then I think

coreywillwhat:
give it a shot now if you don't mind. I have to jump on some calls. If that doesn't work, I can work on it in a few hours.

coreywillwhat:
Tested and the updated code is still working for me.

cdnninja:
> give it a shot again if you have a chance. I have to jump on some calls. If that doesn't work, I can test/work on it in a few hours.

I will leave this one for @Kolbi  to review and merge.

Kolbi:
Tested and working :) Thanks @coreywillwhat

PR #381, opened 2024-04-15
fix: Vehicle Position Timestamp

coreywillwhat:
use `parse_datetime` for the vehicle position timestamp/parktime.


Kolbi:
Tested and working.

PR #382, opened 2024-04-15
fix: Trip Data

coreywillwhat:
### Purpose
This PR brings functionality back to the short term and long term trip data sensors.

### The Issue
The primary issue was the `state` of the sensor returning the full data for the relevant trip which exceeded 255 characters.

### The Solution
State now displays the timestamp from the trip, and attributes have been added to the sensor for the other trip data. Using `get` with defaults to avoid keyerrors.

### Testing
Tested with hardcoded values known to not be available. Tests show the code handled the keyerrors gracefully. Tested with `None` hardcoded as well and 'null' is properly reported.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -337,7 +337,7 @@ class AudiService:
             key=lambda k: k["overallMileage"],
             reverse=True,
         )
-        #_LOGGER.debug("get_tripdata: td_sorted: %s", td_sorted)
+        # _LOGGER.debug("get_tripdata: td_sorted: %s", td_sorted)
         td_current = td_sorted[0]
         _LOGGER.debug("TRIP DATA: td_current: %s", td_current)
         # FIX, TR/2023-03-25: Assign just in case td_sorted contains only one item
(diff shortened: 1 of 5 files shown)


coreywillwhat:
I'm going to add a user option to turn on/off trip data. Just in case this breaks something if someone has a different data structure.

Kolbi:
ShortTerm is working for me, but do you see LongTerm entities?

custom_components/audiconnect/audi_services.py
```diff
@@ -337,10 +337,12 @@ async def get_tripdata(self, vin: str, kind: str):
             key=lambda k: k["overallMileage"],
             reverse=True,
         )
-
+        # _LOGGER.debug("get_tripdata: td_sorted: %s", td_sorted)
         td_current = td_sorted[0]
+        _LOGGER.debug("TRIP DATA: td_current: %s", td_current)
         # FIX, TR/2023-03-25: Assign just in case td_sorted contains only one item
-        td_reset_trip = td_sorted[0]
+        td_reset_trip = td_sorted[1]
```
coreywillwhat, line 344: @Kolbi Can you try changing this line back to:
```python
td_reset_trip = td_sorted[0]
```
This might resolve the long term trip if there is no reset trip

coreywillwhat:
@Kolbi @cdnninja I've converted this to a draft due to @Kolbi 's issue with long term trip data. If changing that line back resolves it. Please feel free to move it back to ready for review, update the PR with that line change and merge it. I probably won't be as active in the coming days/weeks.

Kolbi:
Working now. Merged PR.

Issue #254, opened 2024-03-06
Trip Data Errors


cdnninja:
My car is reporting trip errors.  In my app being a Q7 ICE it doesn't report trips.    As such home assistant throws errors.  I plan to dig in when I get time but this is for a single spot to show this.  Applies to both long and short term trip. 

[...]

PR #388, opened 2024-04-16
fix: Critical Fix - Trip Data 403

coreywillwhat:
going back to #254 it looks like no safeguard was in place for `get_trip_data` and this was part of the reason this was disabled. This means repeated 403 responses with every update if a vehicle isn't reporting trip data.
This PR adds a check similar to the other data functions which will check once, and upon 403 will disable support for the service and wont check again until next HA restart.

### DIFF
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -406,6 +406,7 @@ class AudiConnectVehicle:
         self.support_climater = True
         self.support_preheater = True
         self.support_charger = True
+        self.support_trip_data = True
 
         self.charging_complete_time_frozen = None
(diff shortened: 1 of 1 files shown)


coreywillwhat:
@Kolbi or @cdnninja if you have a few mins, would like to merge this as soon as possible, as well as manually push a new version

Kolbi:
Can test it in around 30 minutes and do the rest. 

Today I had a view in the Audi App with an Q4 etron - the App looks different on that vehicle.

Kolbi:
Tested & working for me.

Let's see on other vehicles :)

coreywillwhat:
Awesome thank you!

PR #389, opened 2024-04-16
fix: More 403 Handling

coreywillwhat:
### Primary Purpose
`refresh_vehicle_data` is currently returning 403. Similar to other functions, like `get_trip_data` and `get_climater` we should check once and disable further tries until a Home Assistant restart. This not only will prevent a user from over polling a forbidden request, it will also increase responsiveness by not waiting for a response each time.

### Other Code Cleanup
- Redact VINs
- Redundancy: if we're trying `update()` (Refresh Cloud Data) regardless of the `refresh_vehicle_data` success or failure, we don't need to include it in each response action. It only needs to be called at the end.
- Exception handling for `refresh_cloud_data`


Kolbi:
Tested & working for me.

Tested the case with ClientResponseError 403 :D

Issue #401, opened 2024-04-18
Sensor `charging_complete_time` unknown on A5


mm98:
**Describe the bug**

First of all. This isn't really a problem at all. I just wanted to report it, since it doesn't really make any sense to me 

I have a `sensor.audi_a5_sportback_charging_complete_time` sensor, which doesn't really have anything to do with my car (gasoline, A5). I can see in the code and debug logs, that the state is checked / validated by `remainingChargingTime` [here](https://github.com/audiconnect/audi_connect_ha/blob/master/custom_components/audiconnect/audi_connect_account.py#L1642-L1670):

```
TRY APPEND STATE: Searching for 'remainingChargingTime' at location=['charging', 'chargingStatus', 'value', 'remainingChargingTimeToComplete_min'], tsoff=-1
TRY APPEND STATE: 'remainingChargingTime' adjusted to 0 due to None value
TRY APPEND STATE: Timestamp for 'remainingChargingTime' is None or missing; not appending state.
```

Wouldn't this mean that it is `None` and returned as `is not supported`? Meaning the sensor would be disabled / hidden?

**Screenshots**
![image](https://github.com/audiconnect/audi_connect_ha/assets/992714/191fdcd7-9af2-4dbc-b4c5-05650cb8df1b)

**Logfile**
[home-assistant_audiconnect_2024-04-18T13-32-10.734Z.log](https://github.com/audiconnect/audi_connect_ha/files/15025688/home-assistant_audiconnect_2024-04-18T13-32-10.734Z.log)

**Your Vehicle Details**
Model: A5 Sportback
Year: 2021
Type (ICE/PHEV/BEV): ICE

**Note:** It might be good to describe what `ICE/PHEV/BEV` is (ex. [this](https://electriccarhome.co.uk/electric-cars/bev-phev-hev-ice/)) for none native English speaking. I had to look it up ;)

Kolbi:
Thanks for reporting. Which version are you using currently?

Good note as I even don't know the term ICE (only in combination with trains :))

Kolbi:
Need to dig deeper but maybe this could be the issue:

`        # Special handling for remainingChargingTime
        if name == "remainingChargingTime" and val is None:
            val = 0
            _LOGGER.debug(
                "TRY APPEND STATE: 'remainingChargingTime' adjusted to 0 due to None value"
            )`

Do you also have the entity remainingChargingTime ?

[...]

PR #402, opened 2024-04-19
fix: charging_complete_time

Kolbi:
Like discussed in https://github.com/audiconnect/audi_connect_ha/issues/401 we always set remaining_charging_time to at least 0 in https://github.com/audiconnect/audi_connect_ha/blob/5055bc34b14e99fda066ce9975da9e8c98ee99da/custom_components/audiconnect/audi_connect_account.py#L1655

### DIFF
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -1652,7 +1652,7 @@ class AudiConnectVehicle:
         if not self.remaining_charging_time_supported:
             return None
         # If there's no last update or remaining time, we can't calculate
-        if self.last_update_time is None or self.remaining_charging_time is None:
+        if self.last_update_time is None or self.remaining_charging_time == 0:
             return None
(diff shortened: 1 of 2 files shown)


Kolbi:
> The current state of this PR would incorrectly make the `charging_complete_time` sensor unavailable for an electric vehicle when the `remaining_charging_time` hits 0. This would remove the sensors ability to state the timestamp in the past, that the vehicle completed charging.
> 
> We should leave this sensor as is, and instead address the `remaining_charging_time` sensor. If remaining charging time isn't available or is unsupported, it should be set to `None` instead of `0`.
> 
> If we use that method, then the charging complete time will not pass the support check and will not display for the user either so solves the gas vehicle issue while not removing functionality of the existing sensors.
> 
> I can work on a PR for this when I have time if thats okay with you @Kolbi

Feel free to contribute to this PR. Hadn't the time to do it.

coreywillwhat:
#408 for this fix

Issue #404, opened 2024-04-19
Add target soc percent "targetSOC_pct" for e-tron


epweston:
First, thank you so much for everyone's work on this project.

I would love to be able to  read the target charge value and was hoping "targetSOC_pct" under "chargingSettings" could be added.
Let me know if I did not supply enough information.

Thank you

Kolbi:
Can you please attach a log file? The return value from the API would be enough if you don't want to attach the whole file. Which etron do you own?

epweston:
Yes, and we have the:

Q4 e-tron® Premium 50 e-tron® quattro® (2023)

[home-assistant_audiconnect_2024-04-19T22-01-02.814Z.log](https://github.com/audiconnect/audi_connect_ha/files/15047113/home-assistant_audiconnect_2024-04-19T22-01-02.814Z.log)

Thanks

Kolbi:
Hi added this PR, do you have the possibility to test it?

mpredfearn:
Maybe it would be worth adding this as a "number" entity in case the API that the API uses to change it becomes known?

Kolbi:
> Maybe it would be worth adding this as a "number" entity in case the API that the App uses to change it becomes known?

Sorry I didn't got your point. :) Can you explain it a little bit more?

mpredfearn:
https://github.com/audiconnect/audi_connect_ha/pull/405 adds "target SoC" as a sensor suggesting it is read only. The app is able to update this, it is read write, so a "number" entity would be more appropriate.

epweston:
> Hi added this PR, do you have the possibility to test it?

Sure np, can I just replace certain files in config\custom_components\audiconnect and restart HA, if so, where to I find the file(s)?
Or do I wait for the releases?
Thanks in advance.

coreywillwhat:
The Target SOC sensor should read the current state of the vehicle as provided by the API. Sensors are by nature unchangeable and are just readings from the device/vehicle/API. I haven't tested, but #405 looks like it meets the requirement of a target SOC sensor.

An `input_number` entity would only make sense as a `control`, but then we would need to figure out the service call to set the target SOC and then update that `input_number` based on the API response and As far as I know, at this time, the service call to set the charger/target SOC of the vehicle is not working due to the API changes and I haven't had time to work on it. I also dont have the ability to set the charger (PHEV) which makes it difficult to test.

[...]

PR #405, opened 2024-04-22
feat: add target SOC

Kolbi:
Added the entity target SOC as asked here: https://github.com/audiconnect/audi_connect_ha/issues/404


coreywillwhat:
I updated the logic to align with `state_of_charge` to use `parse_int`. I've tested the changes and can confirm nothing is broken, but I can't confirm the functionality actually works. Another user will have to let us know.

coreywillwhat (approved):
Reviewed code and tested.
No current functionality is broken. Cannot confirm the new functionality works. Another user will have to confirm that.

PR #411, opened 2024-04-25
fix: return SOC as INT instead of FLOAT

Kolbi:


coreywillwhat:
@Kolbi if you're good with the changes I made, feel free to merge.

coreywillwhat (approved):

PR #415, opened 2024-04-27
fix: Logging: Redact Email Address & Easy (Un)Redact

coreywillwhat:
- Redacts email address in the logs for initial setup and options flow handler.
- Also simplifies the process for controlling log redaction. Users can now toggle log redaction by setting `REDACT_LOGS` in `const.py` to `False` and restart Home Assistant. To enable redaction again, simply set `REDACT_LOGS` back to `True` and restart Home Assistant.


Kolbi:
Cool would be to add it in the https://developers.home-assistant.io/docs/config_entries_options_flow_handler to be able to dis/-enable it without manually changing files.

coreywillwhat:
Good idea. I can work on that. Luckily all the work in this PR is needed anyway for that. Marking as draft for now while I put it together.

PR #422, opened 2024-05-02
fix: Service Endpoints & Code Cleanup

coreywillwhat:
#### US Region
- Execute Vehicle Action
  - Lock/Unlock
  - Climatisation Start/Stop
- Start Climate Control
- Refresh Vehicle Data

#### EU Region
- Refresh Vehicle Data

### Introduce API Levels:
- API Endpoints and data structure are changing for newer vehicles. Currently this must be changed manually in .const file.
- Plan to add under configuration options in future.

### Code Cleanup
- API values in .const

#### Start Climate Control
- Make Logging true to values passed by user
- info level > debug level
- redact vins

### DIFF
--- a/custom_components/audiconnect/audi_account.py
+++ b/custom_components/audiconnect/audi_account.py
@@ -227,16 +227,29 @@ class AudiAccount(AudiConnectObserver):
             await self.connection.set_vehicle_window_heating(vin, False)
 
     async def start_climate_control(self, service):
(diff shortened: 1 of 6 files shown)


coreywillwhat:
Need to troubleshoot re-login during every service call.

coreywillwhat:
Too many changes since this started, will need to start fresh. Saved on my branch for reference if needed.

PR #448, opened 2024-07-24
feat: Radio button for region

cdnninja:
image


cdnninja:
I wouldn't mind someone from EU test this in place as that was the default and optional before. I want to be sure it doesn't break those users.

coreywillwhat:
Tested, working with US

Issue #263, opened 2024-03-07
Exception in appendWindowState + appendDoorState


t0bias-r:
I had an 429 error yesterday (normal for too many requests)
`{"errorCode":"gw.error.VINquotalimitReached","description":"Global quota limit by VIN was reached"}}`
`{"error":{"errorCode":"gw.error.VINquotalimitReached","description":"Global quota limit by VIN was reached"}}`

Further selectiveStatus requests returned and empty Json {}.
`appendWindowState` + `appendDoorState` cause an exception because the existance of the field is not checked:
https://github.com/arjenvrh/audi_connect_ha/blob/8065ff79c462e7daead19338a12f54b08b45f430/custom_components/audiconnect/audi_models.py#L152-L153
https://github.com/arjenvrh/audi_connect_ha/blob/8065ff79c462e7daead19338a12f54b08b45f430/custom_components/audiconnect/audi_models.py#L118-L119

PR #449, opened 2024-07-24
fix: error when window or door data missing

cdnninja:
Potentially solves #263  Someone with actually data for this needs to test though. I don't think my car has this data, errors are gone now for me.


Kolbi:
Tested it on my side and no error occurs. But I didn't had the issue mentioned in #263 

Merged and waiting for tests from @t0bias-r

PR #453, opened 2024-07-24
fix: Disable Services w/ 404 Error

coreywillwhat:
We should disable services that return a 404 error. This will decrease the number of polling errors if an endpoint is dead.

Additional Note: Do we really want 502 to disable services? Couldn't a bad gateway be temporary? This could be the reason GPS locations aren't consistently updated over time until an HA reboot?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -32,5 +32,5 @@ Disable Debug Logging
 **Your Vehicle Details**
 Model:
 Year:
-Type (Combustion/Hybrid/Electric):
+Type (Gas/Hybrid/Electric):
 Region (EU/US/CA/CN):
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -33,4 +33,4 @@ Disable Debug Logging
 Model:
 Year:
 Type (ICE/PHEV/BEV):
-Region (EU/US/CA/CN): 
+Region (EU/US/CA/CN):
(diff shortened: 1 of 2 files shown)


Kolbi:
Need to check because I guess I had some endpoints with 404 in my log and after restart they were working again. But need to check if it really was 404 (wouldnt make sense but I have in my mind that it really was 404).

coreywillwhat:
Yeah let me know what you think. Services are only disabled until next restart when it would check again upon refresh. So I don't think this would cause an issue.

Kolbi:
I just checked my log, I had on the 4. July and 7. July 502 issues with POSITION service. 

Maybe we remove 502 for position for disabling the service?

Or remove 502 in general.... but couldn't find 502 in my log except POSITION

Kolbi:
For climater I also found 403.

Kolbi:
Once also for "ClientResponseError with status 403 while updating trip data"

coreywillwhat:
My understanding is that 502 isn't even reaching the intended server, so it wouldn't cause an issue with over-polling and is often an intermittent issue which may resolve itself.

403 is forbidden, so we should disable if 403 to avoid over-polling a forbidden service.

404 is not found, so we are essentially polling a dead endpoint. We should disable if 404 so we aren't polling a dead endpoint continuously.

let me know if you have other thoughts/ideas.

Kolbi:
I agree with you, so you will remove 502 from disabling the services?

I'm just a little bit confused why I got randolmy an 403 but maybe temporary issues with the Audi server :)

cdnninja:
403 while primarily would occur by design a server outage could cause that in edge causes.   

I'm good with the approach though.

coreywillwhat:
502 Removed. Ready for review.

Issue #450, opened 2024-07-24
`start_climate_control` not working


coreywillwhat:
**Describe the bug**
404 error when running `start_climate_control`

**Your Vehicle Details**
Model: Q5
Year: 2021
Type (ICE/PHEV/BEV): PHEV

Region: US (NA)

pergolafabio:
Hi, i'm posting all sniffed information here: https://github.com/cyr-ius/hass-audiconnect/issues/86

also the stop command for climate is different, and the reposen to get "succeeded" is different

The parking position doesnt work, in the sniff i did from the app, i also get a 404 error using this one:
`https://emea.bff.cariad.digital/vehicle/v1/vehicles/XXX/parkingposition`

t0bias-r:
@coreywillwhat What I do know is that the request structure depends on the functions of the vehicle. For example, it is not possible to configure the temperature or switch on the seat heating in every car. Try a start request with empty request `{}`

coreywillwhat:
> Hi, i'm posting all sniffed information here: [cyr-ius#86](https://github.com/cyr-ius/hass-audiconnect/issues/86)
> 
> also the stop command for climate is different, and the reposen to get "succeeded" is different
> 
> The parking position doesnt work, in the sniff i did from the app, i also get a 404 error using this one: `https://emea.bff.cariad.digital/vehicle/v1/vehicles/XXX/parkingposition`

Interesting...`https://na.bff.cariad.digital/vehicle/v1/vehicles/XXX/parkingposition` (note `na`) is successful for me in the US.

pergolafabio:
SNiffed the temp setting: 

URL:

https://emea.bff.cariad.digital/vehicle/v1/vehicles/XXX/climatisation/settings

Request:

```
{
  "targetTemperature": 21,
  "targetTemperatureUnit": "celsius",
  "climatisationWithoutExternalPower": true,
  "climatizationAtUnlock": false,
  "windowHeatingEnabled": false,
  "zoneFrontLeftEnabled": false,
  "zoneFrontRightEnabled": false
}
```
After fort state if succesfull:

https://emea.bff.cariad.digital/vehicle/v1/vehicles/WAUZZZFZ9RP076552/pendingrequests

response:

```
{
  "data": [
    {
      "id": "ab1cef0e-d047-4fba-9dc3-b3b6691a6336",
      "userId": "0e0f02f7-620c-42e5-87d3-35d76f700a75",
      "vin": "XXXX",
[...]

PR #492, opened 2024-11-12
fix: set_climatisation #450

Kolbi:
[set_climatisation #450](https://github.com/audiconnect/audi_connect_ha/issues/450)

https://github.com/audiconnect/audi_connect_ha/issues/450#issuecomment-2468995526

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -548,7 +548,9 @@ class AudiService:
     async def set_climatisation(self, vin: str, start: bool):
         if start:
             data = '{ "targetTemperature": 23, "targetTemperatureUnit": "celsius", "climatisationWithoutExternalPower": true, "climatizationAtUnlock": false, "windowHeatingEnabled": true, "zoneFrontLeftEnabled": false, "zoneFrontRightEnabled": false, "zoneRearLeftEnabled": false,  "zoneRearRightEnabled": false }'
-            headers = { "Authorization": "Bearer " + self._bearer_token_json["access_token"] }
+            headers = {
+                "Authorization": "Bearer " + self._bearer_token_json["access_token"]
+            }
             res = await self._api.request(
                 "POST",
                 "https://emea.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/start".format(
@@ -556,10 +558,12 @@ class AudiService:
                 ),
                 headers=headers,
                 data=data,
-            ) 
+            )
         else:
-            data = ''
-            headers = { "Authorization": "Bearer " + self._bearer_token_json["access_token"] }
+            data = ""
+            headers = {
+                "Authorization": "Bearer " + self._bearer_token_json["access_token"]
+            }
             res = await self._api.request(
                 "POST",
                 "https://emea.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/stop".format(
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -551,7 +551,7 @@ class AudiService:
             headers = {
                 "Authorization": "Bearer " + self._bearer_token_json["access_token"]
             }
-            res = await self._api.request(
+            await self._api.request(
                 "POST",
                 "https://emea.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/start".format(
                     vin=vin.upper(),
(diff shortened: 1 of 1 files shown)


pergolafabio:
i was testing the start_climate_control with my Q4
do you know why there is a calculation for the temp_c ?
i inspected the json, and it was sending a temp like 25xx , resulting in a bad request...
maybe this was needed for the old API?  i commented it out for me, and below works
again , hardcoded, proably doesnt work for other regions for users with Q4

```
      async def start_climate_control(self, vin: str, temp_f: int, temp_c: int, glass_heating: bool, seat_fl: bool, seat_fr: bool, seat_rl: bool, seat_rr: bool):

        target_temperature = None
        if temp_f is not None:
            target_temperature = int(((temp_f - 32) * (5 / 9)) * 10 + 2731)
        elif temp_c is not None:
            #target_temperature = int(temp_c * 10 + 2731)
            target_temperature = int(temp_c)

        # Default Temp
        target_temperature = target_temperature or 23
        data = { 
            "targetTemperature" : target_temperature, 
            "targetTemperatureUnit": "celsius", 
            "climatisationWithoutExternalPower": True, 
            "climatizationAtUnlock": False, 
            "windowHeatingEnabled": glass_heating, 
            "zoneFrontLeftEnabled": seat_fl, 
            "zoneFrontRightEnabled": seat_fr, 
            "zoneRearLeftEnabled": False, 
            "zoneRearRightEnabled": False 
        }
        data = json.dumps(data)
        headers = {
            "Authorization": "Bearer " + self._bearer_token_json["access_token"]
        }
        res = await self._api.request(
            "POST",
            "https://emea.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/start".format(
                vin=vin.upper(),
            ),
            headers=headers,
            data=data,
        )
```

coreywillwhat:
Q4 data structure is different than all the older vehicles, and as you know, the EU is different than NA region endpoints.

I dont know if the Q4 data structure is going to be the new standard or if only for electric vehicles or some other subset of vehicles. A while back i was working on some logic to use the correct logic based on the vehicle, or even having the user select an option under integration options.

Your Q4 uses Celsius in the new data structure. Legacy data structure used 'deca-kelvin'. Why they used that in the past, i have no idea.

My time is so limited these days, I'm really not sure I'll have time to work on this until the new year, unfortunately.
I could probably get it working pretty quickly for NA/Legacy Data structure, but finding even an afternoon to work on it is hard at the moment. 

Making it work universally and smart enough to do it automatically is the real challenge. I was also working on integration options that the user would select to try different data structures or endpoints. Ultimately that may be the best path forward, but again, last time i worked on it, they changed everything again so I lost the progress.

Happy to answer any questions you have if any others.

pergolafabio:
ok, that makes sence with the temparature
but take your time, no hurry, i have a working solution now, i can turn on/off climate not from HA

niilorm:
> i was testing the start_climate_control with my Q4 do you know why there is a calculation for the temp_c ? i inspected the json, and it was sending a temp like 25xx , resulting in a bad request... maybe this was needed for the old API? i commented it out for me, and below works again , hardcoded, proably doesnt work for other regions for users with Q4
> ...
> 

After some quick and rough testing can confirm this to be working on A3 TFSIe in EU. Will need to be fine tuned if eg. the status needs to be queried. (See [#450](https://github.com/audiconnect/audi_connect_ha/issues/450#issuecomment-2251224138))

Extended the hotfix to enable climatisation to be interrupted as well via _Execute Vehicle Action_. Not pretty as made at 5 AM but it works nevertheless.

audi_services.py
```
async def set_climatisation(self, vin: str, start: bool):
        if start:
            await self.start_climate_control(vin, None, 22, False, False, False, False, False)
        else:

            data = {"": ""}
            data = json.dumps(data)
            headers = {
                "Authorization": "Bearer " + self._bearer_token_json["access_token"]
            }
            res = await self._api.request(
                "POST",
                "https://emea.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/stop".format(
                    vin=vin.upper(),
                ),
                headers=headers,
                data=data,
            )

```

pergolafabio:
Correct, the status is indeed missing, I removed that part , it needs to be included as well, if the status is successful, it wil probably update the climate sensor as well? Now we need to wait for the next 15 min poll interval to actually see if the climate is turned on?

niilorm:
> Correct, the status is indeed missing, I removed that part , it needs to be included as well, if the status is successful, it wil probably update the climate sensor as well? Now we need to wait for the next 15 min poll interval to actually see if the climate is turned on?

Yes, I believe that can be and should be implemented directly via a request to `https://emea.bff.cariad.digital/vehicle/v1/vehicles/{VIN}/pendingrequests`.

However, I don't see it as something really urgent as the action can be assumed successful in most cases.

Will need to be addressed when fixing the API calls.

PR #529, opened 2025-03-13
feat: `Debug Logs` Option

coreywillwhat:
The debug logs option is nice to have for dev work or helping users get more log info quickly. The `update_listener` will be useful for instantaneous changes for the options we currently have, and new options in a future PR. This will be most useful if users need to try multiple options on the fly without having to reboot each time. i.e. trying different endpoints or data format levels.

# Debug Logs Option
### Add Debug Option to Configuration Schema
- **Files Affected:** `__init__.py`
- **Changes:**
  - Updated the integration’s configuration schema to include a new optional boolean key `"debug"` with a default value of `False`.
  - This allows users to enable or disable debug logging via YAML or UI options.

### Set Logger Level at Setup
- **Files Affected:** `__init__.py`
- **Changes:**
  - Modified the `async_setup_entry` function to check for the `"debug"` option.
  - When `"debug"` is enabled, the integration’s logger level is set to `DEBUG`; otherwise, it defaults to `INFO`.

### Update Options Flow to Include Debug Setting
- **Files Affected:** `config_flow.py`
- **Changes:**
  - Added the `"debug"` option to the options flow form so that users can toggle debug mode after initial setup.
  - The options form now displays the current debug setting and allows users to update it.

### Add an Update Listener for Dynamic Debug Changes
- **Files Affected:** `__init__.py`
- **Changes:**
  - Registered an update listener using `config_entry.add_update_listener` in `async_setup_entry`.
  - The listener adjusts both the logger’s level and its handlers’ levels immediately based on the new `"debug"` option.
  - This change ensures that toggling debug mode takes effect immediately without requiring a full reload of the integration.

---

# Code Cleanup
- **Changes:**
  - Changed logging from `info` to `debug`. This is the last usages of `info`. As previously discussed `info` **was** reserved for HA Core, but it seems that guidance has changed to "[Do not print out API keys, tokens, usernames or passwords (even if they are wrong). Be restrictive with _LOGGER.info, use _LOGGER.debug for anything which is not targetting the user.](https://developers.home-assistant.io/docs/development_guidelines/?_highlight=log#log-messages)". 
  - A larger effort will have to take place to remove all usernames/emails/VINs to be sure we are compliant. Additionally, we should be strategic about use of `info` logging.

### DIFF
--- a/custom_components/audiconnect/__init__.py
+++ b/custom_components/audiconnect/__init__.py
@@ -28,9 +28,10 @@ from .const import (
     MIN_UPDATE_INTERVAL,
     RESOURCES,
     COMPONENTS,
+    CONF_DEBUG_LOGS,
 )
 
-_LOGGER = logging.getLogger(__name__)
+_LOGGER = logging.getLogger(DOMAIN)
 
 CONFIG_SCHEMA = vol.Schema(
     {
@@ -53,6 +54,7 @@ CONFIG_SCHEMA = vol.Schema(
                 ),
                 vol.Optional(CONF_REGION): cv.string,
                 vol.Optional(CONF_MUTABLE, default=True): cv.boolean,
+                vol.Optional(CONF_DEBUG_LOGS, default=False): cv.boolean,
             }
         )
     },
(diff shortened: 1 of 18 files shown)


cdnninja:
HA has a built in toggle for this with the logger attribute.  I would think that should be used?

coreywillwhat:
yeah that's good, was just getting my feet wet again, re familiarizing myself with the code and wanted a simple POC for the listener before i work on the other options. Thanks, will close this and work on the others.

PR #530, opened 2025-03-13
fix: Sensors => BinarySensors

Kolbi:
* plug_state
* plug_lock_state


coreywillwhat:
Thank you! Looks good, works well!

PR #531, opened 2025-03-14
fix: Logger.info

coreywillwhat:
## Minor Code Cleanup
- Last of the `_LOGGER.info` changed to `_LOGGER.debug` until we can do a full code cleanup according to [guidelines](https://developers.home-assistant.io/docs/development_guidelines/#log-messages):


cdnninja:
Looks good!

PR #536, opened 2025-03-17
fix: BinarySensors external_power & climatisation_state & charging_state

Kolbi:


Kolbi:
Stupid thing chargingState can have 3 states: charging, readForCharging, notReadyForCharging

coreywillwhat:
I won't be able to review/test until tomorrow, couple initial thoughts:
- For `charging_state` I agree I think an additional sensor would be ideal. Should `charging readiness` be a diagnostic sensor?
- `climatisation_state` can be `["off","cooling","heating"]`. I use these values currently so we need to find a way to capture this data elsewhere if we're changing the sensor. I think the state should be binary with an attribute of heating or cooling. What do you think?

blacknell:
As an FYI, my climatisation state includes `ventilation` as a value

PR #538, opened 2025-03-18
feat: Option - `API Level`

coreywillwhat:
Adds API Levels to the configuration options. This is necessary because the API request data structures differ between vehicles. Users can experiment with different API Levels until they find the one that works for their vehicle. A reference list of supported levels will be maintained in the `README` for easy reference.

*This doesn't do anything yet*


coreywillwhat:
included with #539

PR #542, opened 2025-03-21
fix: Lock/Unlock Service (US)

coreywillwhat:
Fixes the `execute_vehicle_action` > `Lock`/`Unlock`.

_This fix is currently only available in the US. 
We will extend the fix to the EU once we receive input from a user in that region who has remote door lock capabilities and can capture the request_

Issue #540, opened 2025-03-20
404 on lock or unlock


tieskuh:
**Describe the bug**
I am unable to lock and unlock my Audi e-tron 55 from 2022.
The integration works as expected but locking and unlocking doesn't work.

**To Reproduce**

```
action: audiconnect.execute_vehicle_action
data:
  vin: WAUZZZxxxx
  action: unlock
```

**Expected behavior**
The car unlocks

**Logfile**
```
Logger: custom_components.audiconnect.util
Bron: custom_components/audiconnect/util.py:26
integratie: Audi Connect ([documentatie](https://github.com/audiconnect/audi_connect_ha), [problemen](https://github.com/audiconnect/audi_connect_ha/issues))
Eerst voorgekomen: 19 maart 2025 om 23:28:38 (8 gebeurtenissen)
Laatst gelogd: 14:30:27

Unable to lock wauzzzxxxx: 404, message='Not Found', url='https://fal-3a.prd.eu.dp.vwg-connect.com/fs-car/bs/rlu/v1/Audi/DE/vehicles/WAUZZZxxxx/actions'
Unable to unlock wauzzzxxxx: 404, message='Not Found', url='https://fal-3a.prd.eu.dp.vwg-connect.com/fs-car/bs/rlu/v1/Audi/DE/vehicles/WAUZZZxxxx/actions'
```

**Your Vehicle Details**
Model: Audi e-tron
Year: 2022
Type (Gas/Hybrid/Electric): electric
Region (EU/US/CA/CN): EU   (NL)

coreywillwhat:
Locks are next on the list after `start_climate_control`.

tieskuh:
Ah! Thanks!

bla3k:
Here are the details for EU/DE.

I'll post the unlock/lock details in two comments.
Authorization URL to get the securityToken for unlock/lock.

```
GET https://mal-3a.prd.eu.dp.vwg-connect.com/api/rolesrights/authorization/v2/vehicles/WAUZ.../services/rlu_v1/operations/UNLOCK/security-pin-auth-requested

POST https://mal-3a.prd.eu.dp.vwg-connect.com/api/rolesrights/authorization/v2/security-pin-auth-completed
```

### Unlock
`POST https://mal-3a.prd.eu.dp.vwg-connect.com/api/bs/rlu/v1/vehicles/WAUZZ..../unlock`

#### Request
```
accept: application/json
accept-charset: utf-8
[...]

PR #547, opened 2025-03-23
fix: Locking

coreywillwhat:
### Fix for `execute_vehicle_action` > `lock`/`unlock` service call
- Fixes functionality for locking/unlocking for US/EU
_Note: Cloud data may not update in time. Fix coming_

### Code Cleanup
- Adds `host` as optional argument for `_get_vehicle_action_header`

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -460,10 +460,7 @@ class AudiService:
         return body["securityToken"]
 
     def _get_vehicle_action_header(
-        self,
-        content_type: str,
-        security_token: str,
-        host: Optional[str] = None
+        self, content_type: str, security_token: str, host: Optional[str] = None
     ):
         if not host:
             host = (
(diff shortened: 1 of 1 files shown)


coreywillwhat:
@t0bias-r based on your [comment](https://github.com/audiconnect/audi_connect_ha/pull/542#issuecomment-2746103902) i would expect this to work for you. This is working in the US region, unless hashing is different for EU. Let me know if you can test. Thanks!

t0bias-r:
@coreywillwhat: See my comment here https://github.com/audiconnect/audi_connect_ha/pull/542#issuecomment-2747186726

Kolbi:
So combined:
image

@coreywillwhat I'm testing it locally currently and didn't commited it into your PR yet.

coreywillwhat:
Thank you, both! I noticed the issue with the host yesterday. I need to go through and see which (functional) services are using `_get_vehicle_aciton_header` to be sure we won't break something that is working. Also need to test if the `headers` and `data` change will break the US functionality.

Kolbi:
Can confirm Unlock with @t0bias-r  suggestion is working in "DE". But I think the Lock/Unlock status was not updated in time....

coreywillwhat:
Confirmed same functionality can be used for the US.
@Kolbi ready for review.

Kolbi:
We just need to figure which countries expect DE / US are also using this method. This issue is from a user from NL: https://github.com/audiconnect/audi_connect_ha/issues/540

PR #566, opened 2025-03-31
fix: Climate Control: EU/API 0

coreywillwhat:
Fixes the `start climate control` action as well as the execute vehicle action `stop climatisation` for the EU API 0.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -691,7 +691,7 @@ class AudiService:
                     actionid=res["action"]["actionId"],
                 )
             else:
-                headers = self._get_vehicle_action_header("application/json", None)
+                headers = self._get_vehicle_action_header("application/json", None, "msg.volkswagen.de")
                 res = await self._api.request(
                     "POST",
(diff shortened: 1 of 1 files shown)


custom_components/audiconnect/audi_services.py
```diff
@@ -675,6 +675,53 @@ async def start_climate_control(
 
             data = json.dumps(data)
 
+            if country == "US":
+                headers = self._get_vehicle_action_header("application/json", None)
+                res = await self._api.request(
+                    "POST",
+                    "https://mal-3a.prd.eu.dp.vwg-connect.com/api/bs/climatisation/v1/vehicles/{vin}/climater/actions".format(
```
woopstar, line 682: @coreywillwhat the URL contains EU, but the if says the country should be US. 

Doublecheck the URL

coreywillwhat: This is correct for the US endpoint. Confirmed working for me in the US. I agree its confusing that the URL for US has `eu` in it.
The endpoint for EU/API 0 is going to be in the `else:` statement at line 699

PR #609, opened 2025-08-07
feat: add climatisation_at_unlock parameter to climate control

ashleigh-hopkins:
## Description
This PR adds support for the "climatisation at unlock" feature that exists in the native Audi app but was previously unavailable in this integration. The feature allows users to configure whether climate control should continue running when the vehicle is unlocked.

## Problem
Previously, the `climatisationAtUnlock` parameter was hardcoded to `False` in both API level 0 and level 1 implementations, preventing users from keeping climate control active when unlocking their vehicle. This limitation meant the integration couldn't match the full functionality available in the official Audi app.

## Solution
- Added `CONF_CLIMATE_AT_UNLOCK` constant to track the new parameter
- Updated the `start_climate_control` service to accept an optional `climatisation_at_unlock` parameter
- Modified API calls to use the dynamic parameter value instead of hardcoded `False`
- Added the new field to `services.yaml` with a boolean selector for the Home Assistant UI

## Changes Made
- `custom_components/audiconnect/const.py` - Added `CONF_CLIMATE_AT_UNLOCK` constant
- `custom_components/audiconnect/audi_account.py` - Updated service handler to accept and forward the new parameter
- `custom_components/audiconnect/audi_services.py` - Replaced hardcoded `False` values with the configurable parameter in API calls
- `custom_components/audiconnect/services.yaml` - Added UI field for the new parameter

## Testing
- Tested with API level 1 (my vehicle uses this level)
- Verified the parameter is correctly passed through to the API
- Confirmed backward compatibility - defaults to `False` when not specified
- The feature is now available in the Home Assistant UI when calling the service

## Breaking Changes
None - The change is fully backward compatible. If the parameter is not provided, it defaults to `False` (the previous behavior).

## UI Changes
The new option appears in the Home Assistant service UI:
- Field name: `climatisation_at_unlock`
- Description: "Enable climate control to start when vehicle is unlocked"
- Type: Boolean selector

## Additional Context
This feature is particularly useful for users who want to maintain cabin temperature while briefly stepping out of the vehicle, matching the behavior available in the official Audi Connect app.


ashleigh-hopkins:
Hi, I've updated this PR to properly follow the translation structure of the project. I apologize for initially hardcoding the string in services.yaml - I've now moved it to the appropriate translation files (strings.json and en.json). The functionality remains the same, but the implementation now matches the existing pattern for service field translations. Thanks for your patience!

PR #612, opened 2025-08-08
fix: Set suggested display precision to 0 for distance items

rklomp:
This makes distance items like mileage show as rounded number instead of two decimals (which are always `.00` and useless)


rklomp:
I have made a small update and it is tested and working now

cdnninja:
Thank you for this!

Issue #594, opened 2025-05-28
lock not working in v1.13.2 (etron EU)


oleg-d:
**Describe the bug**

Running Execute Vehicle Action lock in EU region does not work anymore. **I was able to find that v1.10.3 it still works**.

v1.13.2 - tried both API 0 and API 1 both the same.

 `Unable to lock wauzzzge: 404, message='Not Found', url='https://mal-3a.prd.eu.dp.vwg-connect.com/api/bs/rlu/v1/vehicles/WAUZZZGE/lock' 
`

**To Reproduce**
Steps to reproduce the behavior: Run  Execute Vehicle Action with action Lock

**Expected behavior**
lock works

**Screenshots**

<img width="1044" alt="Image" src="https://github.com/user-attachments/assets/ce7bbeba-ac9b-4280-9831-17d26d662311" />

**Logfile**
How to enable audiconnect debugging?
Settings > Integrations > Audi Connect > Enable Debug Logging
Run service refresh_cloud_data
Disable Debug Logging

**Your Vehicle Details**
Model: etron
Year: 2020
Type (Gas/Hybrid/Electric): Electric
Region (EU/US/CA/CN): EU

oleg-d:
Finding an old issue https://github.com/audiconnect/audi_connect_ha/issues/496 I rolled back to test 1.10.3 and lock worked for me

pimspierenburg:
Still doesn't seem to work. 

Currently running Audi Connect 1.15, also having an A6 etron in EU, having my Z-Pin also configured. Might this just a certificate that needs to be installed.


Logger: custom_components.audiconnect.util
Bron: custom_components/audiconnect/util.py:26
integratie: Audi Connect (documentatie, problemen)
Eerst voorgekomen: 20:39:23 (3 gebeurtenissen)
Laatst gelogd: 20:39:35

Unable to unlock wauXXXX: Cannot connect to host ha-5a.prd.eu.vwg.vwautocloud.net:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1028)')]
[...]

Issue #564, opened 2025-03-29
SSLCertVerificationError


TS-1900:
Denne fejl stammer fra en brugerdefineret integration.

Logger: custom_components.audiconnect.audi_connect_account
Kilde: custom_components/audiconnect/audi_connect_account.py:223
integration: Audi Connect (dokumentation, problemer)
Første forekomst: 08.39.33 (3 forekomster)
Senest logget: 09.52.15

VEHICLE REFRESH: An unexpected error occurred while refreshing vehicle data for VIN: *************0641. Error: Cannot connect to host ha-5a.prd.eu.vwg.vwautocloud.net:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1028)')]

Kolbi:
Do you still have this issue? Looks like a temporary issue on the Audi server with certificates.

coreywillwhat:
This may have to do with rolling out OAuth as another user suggested a while back.

Is any data retrieved for the vehicle in HA? Any other clues in your logs at what point this occurs? Is login successful, etc. You would have to enable debug logging to see the additional details.

Birkjensen:
Hi, 

If I reload the service, all data is op to date. 
I cant use the Door Lock, it fails. (not response) 
I cant see the charge time and power, if I reload the Audiconnekt app on my phone, it opdate HA, 
Data is only on HA, Monta and AudiConnekt.  


![Image](https://github.com/user-attachments/assets/d83fdef4-13b8-4bd6-b1d4-91c543fc50d6)


Log from HA.. 


2025-04-03 18:35:13.140 ERROR (MainThread) [homeassistant.components.onvif] Failed to migrate unique id for 'camera.hus_frone_substream' as the ONVIF profile index '1' parsed from unique id '44:47:cc:34:fc:ea_1' could not be found
2025-04-03 18:35:32.989 ERROR (MainThread) [custom_components.audiconnect.audi_connect_account] TRIP DATA: An unexpected error occurred while updating trip data for VIN: *************0641. Error: Cannot connect to host ha-5a.prd.eu.vwg.vwautocloud.net:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1028)')]
[...]

Issue #560, opened 2025-03-26
Still encountering 403 error when locking/unlocking in US region after 1.12.2 update


jogan-r:
**Describe the bug**
Still encountering 403 error when locking/unlocking in US region after 1.12.2 update

**To Reproduce**
Lock or Unlock action from any method within the HA website, either via developer tools > action or from the dashboard. Vehicle was added in the US region with API Level 0. I tested changing to API Level 1 with no change. All attributes read correctly, i.e. range, tank level, current lock state, just not able to perform lock or unlock. 

**Expected behavior**
car locks or unlocks respectively

**Screenshots**

**Logfile**
How to enable audiconnect debugging?
Settings > Integrations > Audi Connect > Enable Debug Logging
Run service refresh_cloud_data
Disable Debug Logging

2025-03-26 18:52:31.708 DEBUG (MainThread) [custom_components.audiconnect.audi_account] Retrieved 1 vehicle(s)
2025-03-26 18:52:31.731 DEBUG (MainThread) [custom_components.audiconnect.audi_account] Successfully refreshed cloud data
2025-03-26 18:52:48.722 DEBUG (MainThread) [custom_components.audiconnect.audi_connect_account] Sending command to unlock to vehicle waul2afXXXXXXXXX
2025-03-26 18:52:48.722 DEBUG (MainThread) [custom_components.audiconnect.audi_api] Request initiated: method=GET, url=https://mal-1a.prd.ece.vwg-connect.com/api/rolesrights/authorization/v2/vehicles/WAUL2AFXXXXXXXX/services/rlu_v1/operations/UNLOCK/security-pin-auth-requested, data=None, headers={'User-Agent': 'okhttp/3.7.0', 'X-App-Version': '3.14.0', 'X-App-Name': 'myAudi', 'Accept': 'application/json', 'Authorization': 'Bearer XXX(Imagine a token is here)'}, kwargs={}
[...]

PR #643, opened 2025-09-30
Ignore ssl verification due self-signed certificate in the chain

ArtemKiyashko:
Fix for #564 #594 #560


coreywillwhat:
This has no impact on functionality for me, which is a good thing. Hoping this fixes issues for others. Thanks for this PR!

PR #650, opened 2025-10-05
fix: don't assume a wrong charging_complete_time

neuhausf:
`last_update_time` is not a good assumption for `charging_complete_time`.


coreywillwhat:
Thanks for all of this work! I'll be able to review some tomorrow.

Issue #659, opened 2025-10-24
2021 Q4 Etron will not start climate control from US


evilgeniuschronicles:
**Describe the bug**
The integration is working on the reads correctly. I still get all information in the sensors. However my automation to start climate control on cold days is now failing. It hasn't needed to run in the last 7 months so I don't know when it would have started. I get the error noted in the logfile section (VIN elided). Note it is hitting an URL that says "emea" but I am in the USA.

**To Reproduce**

1. Use the action start_climate_control from the USA

**Expected behavior**
Climate control starts.

**Screenshots**

**Logfile**
```This error originated from a custom integration.

Logger: custom_components.audiconnect.audi_connect_account
Source: custom_components/audiconnect/audi_connect_account.py:358
integration: Audi Connect (documentation, issues)
First occurred: October 23, 2025 at 7:01:01 AM (2 occurrences)
Last logged: 7:01:00 AM

Unable to start climate control of vehicle xxxxxxx6707. Error: 401, message='Unauthorized', url='https://emea.bff.cariad.digital/vehicle/v1/vehicles/XXXXXXXXX6707/climatisation/start'
Traceback (most recent call last):
  File "/config/custom_components/audiconnect/audi_connect_account.py", line 358, in start_climate_control
    await self._audi_service.start_climate_control(
    ...<10 lines>...
    )
  File "/config/custom_components/audiconnect/audi_services.py", line 807, in start_climate_control
    res = await self._api.request(
          ^^^^^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
    )
    ^
  File "/config/custom_components/audiconnect/audi_api.py", line 128, in request
    raise ClientResponseError(
    ...<4 lines>...
    )
aiohttp.client_exceptions.ClientResponseError: 401, message='Unauthorized', url='https://emea.bff.cariad.digital/vehicle/v1/vehicles/XXXXXXXXXX6707/climatisation/start'
```

**Your Vehicle Details**
Model: Q4 Etron
Year: 2021
Type (Gas/Hybrid/Electric): Electric
Region (EU/US/CA/CN):  US

PR #669, opened 2025-11-11
fix: #659 by using existing region formatting logic consistently

evilgeniuschronicles:
Roughly half of the calls to the region specified URLs in this file were using the region formatting code. The other half were just hard-coded to "emea". I replaced all the hard-coded occurrences with the existing logic to make them all consistent. 

Probably even better is to define the region in the initializer but I'm neither good at Python nor do I know any of the design considerations that led to it being like this. I just want automation to once again warm up my car on cold school days.

I replaced this file on my HA instance with my version and I can once again control the climate system from HA.

This addresses #659.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -546,7 +546,6 @@ class AudiService:
             "https://{region}.bff.cariad.digital/vehicle/v1/vehicles/{vin}/charging/mode".format(
                 vin=vin.upper(),
                 region="emea" if self._country.upper() != "US" else "na",
-
             ),
             headers=headers,
             data=data,
@@ -585,7 +584,6 @@ class AudiService:
             "https://{region}.bff.cariad.digital/vehicle/v1/vehicles/{vin}/charging/settings".format(
                 vin=vin.upper(),
                 region="emea" if self._country.upper() != "US" else "na",
-
             ),
             headers=headers,
             data=json.dumps(data),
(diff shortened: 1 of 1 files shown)


custom_components/audiconnect/audi_services.py
```diff
@@ -655,8 +657,9 @@ async def set_climatisation(self, vin: str, start: bool):
                 }
                 res = await self._api.request(
                     "POST",
-                    "https://emea.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/stop".format(
+                    "https://{region}.bff.cariad.digital/vehicle/v1/vehicles/{vin}/climatisation/stop".format(
                         vin=vin.upper(),
+                        region="emea" if self._country.upper() != "US" else "na",
```
cdnninja: What is the intent on this? Right now "region" isn't used as a variable?

cdnninja (changes requested):
Which loop are you referencing?  I don't know this repo code well.

evilgeniuschronicles:
@MaJaHa95  Zero problems from me. I both hate python and am crap at it so there was zero chance I did anything the best way. TBH, once I was asked to justify changes that I barely understand in a weird code file full of gigantically long functions with lots of repeated code, I just lost interest. 

My sole interest is having HA turn on my car's heater. I was trying to be a good person by contributing it back but if I have to, I can just patch my install and fix it. If you can champion this change, do it correctly and navigate the process then I owe you a debt of gratitude.

Kolbi:
Closed as https://github.com/audiconnect/audi_connect_ha/pull/676 was merged.

Issue #668, opened 2025-11-11
Integration won't load


willkpalmer:
**Describe the bug**
Integration won't load on HA Restart or using the reload option

**Screenshots**
![Screenshot_20251111-193152.png](https://github.com/user-attachments/assets/99790cc4-8894-48ca-bca9-0beb411428ef)

**Logfile**
Logger: custom_components.audiconnect.audi_account
Source: custom_components/audiconnect/audi_account.py:208
integration: Audi Connect (documentation, issues)
First occurred: 19:24:13 (2 occurrences)
Last logged: 19:25:32

Failed refresh cloud 

**Your Vehicle Details**
Model: e-tron 55
Year: 2021
Type (Gas/Hybrid/Electric): Electric
Region (EU/US/CA/CN): EU

sspierenburg:
Most probably this is caused by rate limiting. I have seen the same problem and I’m having problems with the app too and logging into the main user in the car is also not working. I’ve disabled the integration temporarily to check if the account will recover itself

sspierenburg:
I can confirm everything recovered, so it's rate limiting. The question whether they limit on amount of requests or fingerprinting remains.

chrisrbk86:
I get this problem from time to time, and I’m currently facing it now. 

What’s the fix, to disable it? How long for? I tried 12 hours and it still doesn’t work. 

Any suggestions? Thanks.

PR #675, opened 2025-11-29
feat: Add option to exclude specific VINs to prevent rate limiting

breinkober:
This update introduces a new configuration option to filter out specific VINs from being queried. This addresses an issue where querying retired or problematic vehicles (e.g., old leases still linked to the account) on deprecated endpoints could trigger API rate limits and temporary account bans. #668

Changes:
- Added `CONF_FILTER_VINS` constant.
- Updated `config_flow.py` to include a "Filter VINs" field in the integration options (comma-separated list).
- Updated `AudiConnectAccount` to accept and store excluded VINs.
- Modified `AudiConnectAccount.update` to skip vehicle discovery and updates for VINs in the exclusion list.
- Updated `audi_account.py` to pass the configured exclusion list during initialization.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Gets semantic release info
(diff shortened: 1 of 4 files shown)


Kolbi:
Merged. String & Translations are missing.

PR #680, opened 2025-12-15
fix: Fix pre-heater endpoint

majd:
This pull request contains a small fix to a URL path typo in the `set_pre_heater` method. The change corrects the spelling of "auxilaryheating" to "auxiliaryheating" in the API endpoint, ensuring requests are made to the correct URL. The regression was introduced in #676.


Kolbi:
> @Kolbi Can we get a hotfix out with this change?

Thanks for applying hotfix :) Merged and new version released.

PR #749, opened 2026-06-02
fix: handle missing refresh_token in mbboauth auth response (fixes #748)

### LINKED ISSUE #748: Invalid credentials
**Describe the bug**
The latest update (v2.1.0) failed to initialize after HA restart. Reconfiguring reported "Invalid credentials". I have since tried different versions (1.19.2, 2.0.1) but the issue is the same. It was working before the last upgrade.

**To Reproduce**

1. Go to 'Settings'
2. Click on 'Devices & Services'
3. Click on '+ Add integration' and select Audi Connect
4. Enter username + password
5. Click 'Submit'
6. See "Invalid credentials"

**Expected behavior**
To be logged in

**Screenshots**

**Logfile**
How to enable audiconnect debugging?
Settings > Integrations > Audi Connect > Enable Debug Logging
Run service refresh_cloud_data
Disable Debug Logging

**Your Vehicle Details**
Model: Q4 45 e-tron Sportback
Year: 2026
Region (EU/US/CA/CN): EU

charliegadget:
## Problem

After a recent Audi/Cariad backend change, the MBB OAuth `/mobile/oauth2/v1/token` endpoint (`grant_type: id_token`) no longer returns a `refresh_token` in its initial auth response.

The login flow crashes at the final step:

```
KeyError: 'refresh_token'
File "audi_services.py", line 1464, in login_request
    "token": mbboauth_auth_json["refresh_token"],
```

This happens **after** all authentication steps complete successfully, so credentials are entirely valid — but the error surfaces as "Invalid Credentials". This is the root cause of #748.

## Fix

Make the immediate token refresh conditional on `refresh_token` being present in the auth response. If absent, use the auth token directly as `vwToken`.

The `refresh_token_if_necessary` method already handles the no-refresh-token case correctly (it guards with `if "refresh_token" not in self.mbboauthToken: return False`), so ongoing token renewal is unaffected.

## Testing

Confirmed working on v2.1.0, UK/EMEA region, Audi e-tron. Integration authenticates and loads all vehicle data successfully after this fix.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -1499,4 +1499,3 @@ class AudiService:
 
 
 __all__ = ["AudiService"]
-
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -1452,33 +1452,41 @@ class AudiService:
         self.mbboauthToken = mbboauth_auth_json
 
         # mbboauth refresh (app immediately refreshes the token)
-        headers = {
-            "Accept": "application/json",
-            "Accept-Charset": "utf-8",
-            "User-Agent": AudiAPI.HDR_USER_AGENT,
-            "Content-Type": "application/x-www-form-urlencoded",
-            "X-Client-ID": self.xclientId,
-        }
-        mbboauth_refresh_data = {
-            "grant_type": "refresh_token",
-            "token": mbboauth_auth_json["refresh_token"],
-            "scope": "sc2:fal",
-            # "vin": vin,  << App uses a dedicated VIN here, but it works without, don't know
-        }
-        encoded_mbboauth_refresh_data = urlencode(
-            mbboauth_refresh_data, encoding="utf-8"
-        ).replace("+", "%20")
-        mbboauth_refresh_rsp, mbboauth_refresh_rsptxt = await self._api.request(
-            "POST",
-            self.mbbOAuthBaseURL + "/mobile/oauth2/v1/token",
-            encoded_mbboauth_refresh_data,
-            headers=headers,
-            allow_redirects=False,
-            cookies=mbboauth_client_reg_rsp.cookies,
-            rsp_wtxt=True,
-        )
-        # this code is the old "vwToken"
-        self.vwToken = json.loads(mbboauth_refresh_rsptxt)
+        # The MBB OAuth auth response no longer always includes a refresh_token.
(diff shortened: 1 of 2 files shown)


woopstar:
Confirmed working for me in Denmark with my Audi Etron from 2019 using the DE region.

mm98:
Also working here in Denmark, with Audi A5 (9.5) from 2021 using DE region.

vla89:
Confirmed working here also, DE region, Q4 e-tron.

n0llp0inter:
Thanks for the fix! Confirmed to work on PPE platform as well (Q6)

danielcolquitt:
Confirmed working for me — thank you for the fix.

Setup:

* Audi Connect v2.1.0
* UK / EMEA account
* Audi Q8 50 e-tron
* Home Assistant Container running on a Raspberry Pi 4B
* API level 1
* Integration used with custom Home Assistant automations for Audi/Ohme SoC synchronisation, target SoC setting and climate-control actions

Before applying this patch, authentication failed when setting up/reloading the v2.1.0 integration because the auth response did not contain a refresh_token.

I applied the change from this PR to custom_components/audiconnect/audi_services.py, restarted Home Assistant, and the integration authenticated successfully. My vehicle entities are now available again and the integration is working with my existing automations using the new v2 device_id service calls.

Thanks for diagnosing and fixing this.

charliegadget:
Thanks for the suggestion, but I think this has a semantic issue. The
grant_type: id_token exchange is designed to accept an OIDC id_token (which
is what the earlier mbboauth_auth_data step already uses —
self._bearer_token_json["id_token"]). Passing an access_token to that grant
type is non-standard and untested against the MBB OAuth endpoint.

The key insight is that by the time we reach the refresh step, we already
have a fully authenticated session — mbboauth_auth_json contains a valid
access_token that works as vwToken directly. The immediate refresh is just
mimicking what the myAudi app does, but it isn't required for the
integration to work. Skipping it when refresh_token is absent is simpler,
avoids an extra network call, and has been confirmed working.

If you've actually tested the id_token + access_token fallback against the
live MBB endpoint and it succeeds, happy to reconsider — but it would need
verification first.

charliegadget:
Thanks for the pre-commit.ci auto-fixes — trailing newline in `audi_services.py` and the readme indentation are now sorted. Great to see so many confirmations across different regions and vehicles. Happy for the maintainer to merge when ready.

coreywillwhat:
I dont see any responses from the US as we're all just waking up to this. I'll test and merge if no issues.

PR #750, opened 2026-06-02
feat!: add charging profiles

vla89:


vla89:
This was an error, was supposed to merge within own fork.

PR #751, opened 2026-06-02
feat: added climate control entity

lukasl96:


coreywillwhat:
Is this ready for review or still in draft? If ready, what testing has been done?

its-me-prash (commented):
Had a look at this since it has been sitting a while. A climate entity is a good idea and a much nicer fit than the service call for anyone driving this from a thermostat card. A few things stand out that would stop it behaving as intended today.

**The entity never reflects the car's state.** `attr_key` is `climatisation_active`, but the vehicle property is called `climatisation_state`, and there is already a sensor using it in `sensor.py`. Since `getattr` only falls back when the attribute is missing entirely, `hvac_mode` and `hvac_action` always end up reading `self._is_on`, which is whatever Home Assistant last set locally. After a restart it reports off regardless of what the car is actually doing.

Worth knowing that `climatisation_state` holds a string from the climater payload rather than a bool, so swapping the key alone is not enough. A value of `off` is still truthy.

**`current_temperature` is always None.** There is no `temperature` property on the vehicle. The nearest is `outdoor_temperature`, but that is the outside reading rather than a cabin temperature, so leaving `current_temperature` unimplemented is probably more honest than wiring it to that.

**No support gating.** The other platforms check `is_entity_supported(vehicle, description.attr_key)` before appending, so a car without the feature does not get a dead entity. This adds one for every vehicle unconditionally.

**Half degree steps do not survive the trip.** `_attr_target_temperature_step` is 0.5, but `_start_climate_control` converts to Fahrenheit with `int()` and the service layer converts back with `int()` again, so 21.5 reaches the car as 21. Both `temp_f` and `temp_c` are passed and the service layer prefers `temp_f`, so the lossy path is always the one taken. Passing only `temp_c` removes one rounding step, though the service layer still does `int(temp_c)`, so half degrees are not representable on API level 1 as things stand.

**This would now always send a full settings payload.** #822 went in earlier today: on API level 1 the body is omitted when the caller supplied no settings, because some cars reject a settings payload outright, which is the Q6 e-tron problem in #771. `_start_climate_control` fills every field with a default, so it always produces a full body and those cars would keep failing through this entity. Forwarding only what the user actually set would let them work.

Two smaller ones: `duration` is put into `params` at both call sites but never read by `_start_climate_control`, and the name is hardcoded to `Klimatisierung` while every other description in the repo is English.

A rebase would be worth it too. This is based on master from early June and a fair amount has moved since, authentication included. Also note the target branch is `beta` now rather than master.

Kolbi:
I close the PR as it was implemented with: https://github.com/audiconnect/audi_connect_ha/pull/849

Thanks for your help @lukasl96

PR #754, opened 2026-06-05
feat: Restore API requests remaining sensor state

### LINKED ISSUE #753: Restore last known value of "API requests remaining" sensor after Home Assistant restart
**Is your feature request related to a problem? Please describe.**

The new `API requests remaining` diagnostic sensor is very useful for monitoring the apparent remote-operation allowance reported through `Vcf-Remaining-Calls`.

At present, the sensor becomes `unknown` after a Home Assistant restart and remains unknown until an Audi response containing that header is received. This is inconvenient because the last reported allowance is operationally useful, and obtaining a fresh value may itself require performing a remote vehicle operation.

In my case, this matters because I use Audi Connect with Home Assistant automations for an Audi Q8 e-tron, including SoC synchronisation with an Ohme charger and climate-control actions. I am deliberately reducing automated vehicle refreshes after observing that the allowance appears to be shared with normal myAudi use: after Home Assistant reported 49 remaining calls, I stopped climate control using the official myAudi app and then ran one Home Assistant vehicle/SoC refresh; the next value reported in Home Assistant was 47. I appreciate that this observation does not establish the precise quota scope, but it does make retaining the most recently observed value particularly useful.

**Describe the solution you'd like**

Please consider restoring the last known numeric value of the existing `API requests remaining` diagnostic sensor after a Home Assistant restart, until a fresh `Vcf-Remaining-Calls` value is returned by Audi.

Because a restored value may no longer be current — for example, calls may have been made through the myAudi app while Home Assistant was offline, or the allowance may have reset — it would be helpful for the sensor also to expose freshness information, such as:

- a `last_observed` timestamp recording when the value was last received from Audi; and/or
- an indication that the currently displayed value was restored rather than freshly reported during the present runtime.

Conceptually, the sensor could use Home Assistant's restore-state support, while preserving the distinction between "last known value" and "freshly observed value".

I would also be happy to contribute a PR implementing this approach if it would be useful.

**Describe alternatives you've considered**

I can work around this locally by creating a trigger-based template sensor that mirrors `sensor.<vehicle>_api_requests_remaining` only when it has a numeric state and restores the last valid value over Home Assistant restarts.

That works for an individual setup, but it seems preferable for the integration-provided diagnostic entity to retain this information itself, because:

- the rate-limit sensor is specifically intended to help users avoid exhausting an Audi allowance;
- every user currently has to create their own persistence workaround;
- the integration is best placed to expose whether a value is restored or freshly received from Audi.

I do not think silently restoring the value without a timestamp or freshness indication would be ideal, since the quota appears capable of changing outside Home Assistant.

**Additional context**

Setup in which I observed this behaviour:

- Audi Connect v2.1.0, with the authentication fix from PR #749 applied locally
- I have also checked the v2.1.2 source, where the relevant sensor/API behaviour appears unchanged
- UK / EMEA Audi account
- Audi Q8 50 e-tron
- API level 1
- Home Assistant Container running on a Raspberry Pi 4B
- Audi Connect used alongside an Ohme charger integration for SoC synchronisation and preconditioning actions

From reading the v2 source, `vcf_remaining_calls` is initialised as `None` when the API object is created, and the `API requests remaining` sensor exposes that in-memory value directly. That appears to explain why the diagnostic entity loses its state after a Home Assistant restart until another qualifying Audi response is received.

**Logfile**

No error logfile is associated with this request: the sensor is functioning as implemented, but loses its last known value after a Home Assistant restart.

Reproduction steps:

1. Perform an Audi operation that causes `API requests remaining` to show a numeric value.
2. Restart Home Assistant.
3. Observe that `API requests remaining` becomes `unknown`.
4. Perform another qualifying Audi operation.
5. Observe that the sensor receives a numeric value again.

**Your Vehicle Details**

Model: Audi Q8 50 e-tron  
Year: 2024  
Type (ICE/PHEV/BEV): BEV

danielcolquitt:
## Summary

Restore the last known value of the `API Requests Remaining` diagnostic sensor after a Home Assistant restart, and expose when the value was last genuinely observed from Audi.

Fixes #753

## Background

The `API Requests Remaining` sensor is useful for avoiding unnecessary vehicle refreshes and other remote operations. At present, the value is held only in memory, so after a Home Assistant restart the sensor becomes `unknown` until Audi next returns a `Vcf-Remaining-Calls` header.

That is inconvenient because obtaining a fresh counter value may itself require a quota-bearing Audi operation.

This PR keeps the existing diagnostic sensor, but makes its last known value restorable after restart and adds metadata to distinguish a restored value from a value freshly observed in the current Home Assistant runtime.

## Changes

* Record `vcf_remaining_calls_last_observed` when Audi returns a valid `Vcf-Remaining-Calls` header.
* Convert `AudiApiRateLimitSensor` to use Home Assistant's sensor restore support.
* Restore the last known numeric API request counter after Home Assistant restart.
* Add sensor attributes:

  * `last_observed`: timestamp at which Audi last supplied a `Vcf-Remaining-Calls` value;
  * `value_restored`: `true` when the displayed value was restored from the previous Home Assistant runtime and has not yet been replaced by a fresh observation.
* Document the restored-value semantics in the README.

## Rationale

A restored value is useful, but it should not be treated as definitely current. The remaining allowance may have changed outside Home Assistant, for example because of actions performed through the official myAudi app or because of quota replenishment/reset behaviour.

The `last_observed` and `value_restored` attributes make that distinction explicit:

* `value_restored: false` means the value was observed from Audi in the current runtime;
* `value_restored: true` means the value was restored from Home Assistant's previous state and may no longer represent the current allowance.

## Attribute naming

The restored-state flag is named `value_restored` rather than `restored`.

During local testing, using an attribute named `restored` caused Home Assistant's frontend to display the restored/orphaned-entity warning:

> This entity is no longer being provided by the integration.

This appears to be because Home Assistant's frontend treats a truthy `stateObj.attributes.restored` value as indicating a restored entity that is no longer being provided by an integration. See Home Assistant frontend issue:

https://github.com/home-assistant/frontend/issues/4401

Using `value_restored` avoids colliding with that frontend convention while still making the sensor semantics clear.

## Notes on quota behaviour

I have also been observing the counter behaviour while using Audi Connect v2.1.2.

A `refresh_cloud_data` action successfully performed a cloud update, but did not return a new `Vcf-Remaining-Calls` observation. The sensor value was re-reported by Home Assistant with the previously cached value, but no fresh quota observation was received.

I also have preliminary evidence that the counter may be shared with actions performed through the official myAudi app, but I have not attempted to document or model that in this PR. This PR intentionally does not add optimistic replenishment logic or broaden the documentation for quota-consuming service actions. Those would be better handled separately once the behaviour has been observed more rigorously.

## Testing

Tested locally with:

* Audi Connect v2.1.2
* Home Assistant Container
* UK / EMEA Audi account
* Audi Q8 50 e-tron
* API level 1

Local checks:

* Ran `python -m compileall custom_components/audiconnect`
* Ran `pre-commit run --files custom_components/audiconnect/audi_api.py custom_components/audiconnect/sensor.py readme.md`

Functional testing in Home Assistant Container:

1. Installed the patched `audi_api.py` and `sensor.py`.
2. Performed one `refresh_vehicle_data` action.
3. Confirmed `API Requests Remaining` reported a numeric value with:

   * `value_restored: false`
   * populated `last_observed`
4. Restarted Home Assistant.
5. Confirmed the same numeric value was restored with:

   * `value_restored: true`
   * previous `last_observed` preserved
6. Confirmed that using the attribute name `restored` caused Home Assistant's restored/orphaned-entity warning, and that renaming the attribute to `value_restored` avoided that warning.
7. Confirmed `refresh_cloud_data` did not falsely clear the restored-state metadata when Audi did not return a new `Vcf-Remaining-Calls` value.

### DIFF
--- a/custom_components/audiconnect/audi_api.py
+++ b/custom_components/audiconnect/audi_api.py
@@ -4,7 +4,7 @@ import asyncio
 import json
 import logging
 from asyncio import CancelledError, TimeoutError
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Any
 
 from aiohttp import ClientResponseError, ClientSession
@@ -33,6 +33,7 @@ class AudiAPI:
             {"http": proxy, "https": proxy} if proxy else None
         )
         self.vcf_remaining_calls: int | None = None
+        self.vcf_remaining_calls_last_observed: datetime | None = None
 
     def use_token(self, token: dict[str, Any] | None) -> None:
         self.__token = token
@@ -83,6 +84,7 @@ class AudiAPI:
                             remaining = None
                         if remaining is not None:
                             self.vcf_remaining_calls = remaining
+                            self.vcf_remaining_calls_last_observed = datetime.now(UTC)
                             if remaining < 10:
                                 _LOGGER.warning(
                                     "[VCF-RATE-LIMIT] Remaining calls: %d — warning: VCF call limit nearly exhausted",
--- a/custom_components/audiconnect/sensor.py
+++ b/custom_components/audiconnect/sensor.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 from typing import Any
 
 from homeassistant.components.sensor import (
+    RestoreSensor,
     SensorDeviceClass,
     SensorEntity,
     SensorEntityDescription,
@@ -457,7 +458,7 @@ class AudiSensor(AudiEntity, SensorEntity):
         return None
 
 
-class AudiApiRateLimitSensor(AudiEntity, SensorEntity):
+class AudiApiRateLimitSensor(AudiEntity, RestoreSensor):
     """Account-level sensor exposing the Vcf-Remaining-Calls API rate limit."""
 
     _attr_name = "API requests remaining"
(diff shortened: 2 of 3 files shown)


Kolbi:
As I am not able to test it, can anyone confirm this to be working?

danielcolquitt:
I can confirm that it's been working for me for around a month now.

PR #770, opened 2026-07-09
fix: update `plug_state` to account for `invalid` state in home zone

### LINKED ISSUE #760: Charging related states incorrect
My Audi A6 e-tron (early version from summer 2025) states wrong charging related states (sensor.charging_state/plug state) as soon as it is at home.

It shows "plugged" although the cable is not connected/plugged at all. As soon as I leave the home zone, it switches back to "not plugged".

Any one else experiencing this behavior.

It's now since approx. 2 month.

Not sure if it is related to issue #713 somehow.

coreywillwhat:
In the home zone, the vehicle can sometimes return `invalid` state when not plugged in. When leaving the home zone the vehicle properly reports `disconnected`.

This fix will set the state of the `Plug State` sensor to `Unplugged` if in `invalid` state.

### DIFF
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -1905,7 +1905,7 @@ class AudiConnectVehicle:
         """Return plug state"""
         if self.plug_state_supported:
             check = self._vehicle.state.get("plugState")
-            return check != "disconnected"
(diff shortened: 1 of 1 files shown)


coreywillwhat:
Tested by @manu240429 https://github.com/audiconnect/audi_connect_ha/issues/760#issuecomment-4943933977

Issue #756, opened 2026-06-10
Setup Error: Login to Audi service failed: 'access_token' during re-authentication


RajAdmin83:
Description:

Describe the bug

The integration had been working perfectly up until yesterday. After adjusting some charging timers in my official myAudi mobile app, the Home Assistant integration dropped offline. When attempting to reload, I was met with authentication errors.

I completely removed the integration, uninstalled the component from HACS, restarted Home Assistant, and attempted a fresh setup. However, I am completely blocked at the initial setup configuration screen. When entering my correct myAudi credentials, it fails immediately with an "Invalid credentials" error in the UI.

Checking my core logs, the backend error states:

LOGIN: Failed to log in to the Audi service: 'access_token'. You may need to open the myAudi app, or log in via a web browser, to accept updated terms and conditions.

I have verified my credentials are 100% correct, and I can log into both the official myAudi smartphone app and the web portal without being prompted for any new terms or conditions.

To Reproduce
	1.	Delete existing Audi Connect integration configuration.
	2.	Restart Home Assistant.
	3.	Go to Settings > Devices & Services > Add Integration.
	4.	Search for "Audi Connect" and enter username/password.
	5.	See error: "Invalid credentials" in the UI.

Expected behavior
The integration should successfully complete the OpenID/OAuth handshake, retrieve the access token from the VW Group/Cariad authentication endpoints, and initialize the vehicle entities.

Logfile

Logger: custom_components.audiconnect.audi_connect_account
Source: custom_components/audiconnect/audi_connect_account.py
Integration: Audi Connect
ERROR - Login to Audi service failed: 'access_token'.

Your Vehicle Details
Model: Audi Q4 e-tron (2026)
Region: UK (Europe)

UptimizeNL:
I'm seeing the exact same thing here. Tried everything, logging in to the app and website, expected to find some ToS I have to accept, but nothing... Did you manage to find a solution?

RajAdmin83:
I got it working by leaving long periods (8+ hours) before attempting a fresh install. Has been working for a few weeks without issue - though I did have API refresh set to 60 instead of the deault 15 - unclear whether that had any effect but I do not need refreshes that regularly. The integration went down randomly yesterday afternoon - no interaction from my end. Multiple login attempts unsuccessful. Tried removing the integration and HACS instance and then reinstall with the HA restart before signing into the integration - all with no luck.

[...]

PR #777, opened 2026-07-15
feat!: device-code login to fix auth broken by Play Integrity attestation

### LINKED ISSUE #772: Login fails with KeyError('access_token') — CARIAD token endpoint returns 400 "invalid assertion headers"; hybrid flow blocked by "invalid client"
**Describe the bug**

Since early July 2026 the integration can no longer log in. Every attempt ends with:

```
LOGIN: Failed to log in to the Audi service: 'access_token'
```

That `KeyError('access_token')` is misleading — the myAudi app and web portal work fine, so credentials and consent are not the problem. Adding temporary logging around the token exchange shows the real cause: **VW/CARIAD now rejects the authorization-code → token exchange** at `/auth/v1/idk/oidc/token` with `400 {"error":"invalid assertion headers"}`, so the response contains no `access_token`. The hybrid-flow workaround from robinostlund/volkswagencarnet#333 (`response_type=code id_token token`) is **also** rejected for the audi client, with `400 {"error":"invalid_request","error_description":"invalid client"}`. Both auth paths are currently blocked (full captures in the Logfile section).

**To Reproduce**

1. Configure the integration (EU region) with valid myAudi credentials.
2. Restart Home Assistant (or reload the integration / call the `refresh_cloud_data` service).
3. See the error: repeated `LOGIN: Failed to log in to the Audi service: 'access_token'`, integration stays in `setup_retry`, all entities `unavailable`.

**Expected behavior**

Integration logs in successfully and refreshes cloud data.

**Screenshots**

N/A — backend/auth issue; the captured HTTP responses are in the Logfile section.

**Logfile**

The standard debug log only shows the misleading `KeyError('access_token')`. To find the real cause I added temporary `_LOGGER.warning` lines around (a) the bearer-token `POST` in `login_request()` and (b) the initial authorize `GET`, to capture the actual HTTP status + body.

*1. Current code path — `authorization_code` → token exchange is rejected:*

```
POST .../auth/v1/idk/oidc/token      (grant_type=authorization_code, PKCE code_verifier)
→ HTTP 400
{"error":"invalid assertion headers"}
```

`json.loads()` succeeds but the dict is only `{"error": ...}`, so `self._bearer_token_json["access_token"]` raises `KeyError('access_token')`. This matches the CARIAD BFF change described in robinostlund/volkswagencarnet#333 (auth code is now bound to the CARIAD BFF as the sole authorised exchanger; direct exchange from the public client is refused).

*2. Hybrid-flow workaround (PR #333 approach) — rejected for the audi client:*

```
GET https://identity.vwgroup.io/oidc/v1/authorize
    ?response_type=code+id_token+token
    &client_id=09b6cbec-cd19-4589-82fd-363dfa8c24da@apps_vw-dilab_com
    &redirect_uri=myaudi:///&code_challenge=...&code_challenge_method=S256&prompt=login&scope=...openid
→ HTTP 400
{"error":"invalid_request","error_description":"invalid client"}
```

The **same `client_id` works with `response_type=code`**, but is rejected as *"invalid client"* the moment `id_token token` is added — i.e. this client is registered for authorization-code flow only and is not permitted to use the implicit/hybrid response types. So PR #333's fix does not port directly (the VW client it targets permits hybrid; the audi client does not).

Summary — both paths blocked:

| Approach | Result |
|---|---|
| `grant_type=authorization_code` → token endpoint | `400 {"error":"invalid assertion headers"}` (CARIAD BFF) |
| `response_type=code id_token token` (hybrid) | `400 {"error":"invalid_request","error_description":"invalid client"}` (Auth0) |

A working fix likely needs the correct `client_id` and/or flow that the current myAudi app uses against the CARIAD backend. Endpoint URLs are already migrated to `/auth/v1/idk/oidc/*` in 2.1.4, so that part is fine.

Version note: reproduced on **audi_connect_ha 2.1.4** and **2.2.0** — the auth code in `audi_services.py` / `audi_connect_account.py` is unchanged between the two, so upgrading does not help. Home Assistant 2026.7.x. Related: #769 (API down?!).

**Your Vehicle Details**
Model: e-tron 55 Sportback (Black Edition)
Year: 2022
Region (EU/US/CA/CN): EU

### LINKED ISSUE #775: Login fails with "Invalid credentials" although myAudi credentials work in app and browser
## Description & ## Environment

Home Assistant version:
Core 2026.7.2
Supervisor 2026.06.2
Operating System 18.1
Frontend 20260624.3

Audi Connect integration version:  V 2.2.0

Browser login on my.audi.com working: Yes, is working without problems.

Audi Connect integration version
Vehicle type/model/year
Browser login on my.audi.com working: Yes/No

The Audi Connect integration no longer accepts my login credentials. Home Assistant shows the error:

`Invalid credentials`

However, the same credentials work correctly in the official myAudi app. I also changed the password in myAudi and verified that I can still log in successfully with the new password outside of Home Assistant.

The integration still rejects the login with the same "Invalid credentials" message.

## What I tried

- Verified the username and password in the official myAudi app
- Changed the myAudi password
- Logged in again successfully in the official myAudi app
- Retried the Audi Connect integration setup in Home Assistant
- Used region `DE`
- Used a 15 minute polling interval
- Restarted Home Assistant
- Tried re-entering the credentials several times

The issue persists.

## Expected behavior

The integration should accept the valid myAudi credentials and complete the setup/login process.

## Actual behavior

The integration immediately fails with:

`Invalid credentials`

## Screenshot

The setup dialog shows the message "Ungültige Anmeldeinformationen" / "Invalid credentials" although the credentials are valid.

## Additional context

This does not look like a normal wrong-password issue because the same login works in the official myAudi app. It may be related to a recent Audi/CARIAD login or token endpoint change.

Please let me know if there is anything specific I should test, such as a different API level, additional debug logging, or a specific branch/version.

jonny190:
## What this fixes

Since early July 2026 login has failed with `KeyError('access_token')` (issues #772, #775, #756). The real failure is the final token exchange:

```
POST https://emea.bff.cariad.digital/auth/v1/idk/oidc/token  (grant_type=authorization_code)
-> 400 {"error":"invalid assertion headers"}
```

Audi/CARIAD now require a genuine Google Play Integrity attestation (an `X-Assertion` header) on the authorization-code token exchange. The current myAudi app carries a full Play Integrity stack for this (`technology.cariad.cat.playintegrity.*`, `IdentityKit$Error$PlayIntegrity`). That token cannot be produced off-device, so the username and password flow this integration used can no longer complete, and the stale `X-QMAuth` header it sent is rejected.

Switching to a hybrid OIDC flow (as the VW and Cupra connectors did) does not work for Audi: the Audi client is registered for the basic authorization-code flow only, and every `response_type` that returns a front-channel token is rejected with `invalid client` (verified live against `identity.vwgroup.io`).

## The approach

Use the OAuth Device Authorization Grant (RFC 8628) instead. The same CARIAD token endpoint accepts the `device_code` grant with no attestation header, and the refresh-token grant is un-attested too. This was verified end to end against a live account: the device flow yields a token, that token is accepted by the vehicle API (`GET /vehicle/v1/vehicles` returns the vehicle), and the refreshed token keeps working. It has also been tested in a running Home Assistant instance.

Sign-in becomes a one-time browser approval (the myAudi device-code page), after which the integration stores the IDK refresh token and refreshes headlessly, exactly as it did before.

## Changes

- `audi_services.py`: add `request_device_code`, `poll_device_token` and `login_with_refresh_token`; share the AZS and mbboauth token derivation via `_finalize_session`; drop the password and HTML-scrape flow and the stale `X-QMAuth` header; raise a typed auth error when a token is rejected.
- Account layer: token-based auth, persist rotated refresh tokens back to the config entry, and raise a dedicated auth error so Home Assistant starts reauth when the stored token expires.
- Config flow: device-code setup and reauth steps; reconfigure keeps region, S-PIN and API level.
- Migrate config entries v2 to v3 (removes stored credentials and prompts a one-time device-code sign-in).

## Breaking change

Existing entries migrate to v3 and must sign in again through the new device-code flow. The username and password are no longer used or stored.

## Notes

Fixes #772. Likely also resolves #775 and #756.

This was investigated and implemented with Claude Code, including live verification of the device-code flow and an adversarial review pass. Happy to adjust anything to fit the project's preferences.

### DIFF
--- a/custom_components/audiconnect/__init__.py
+++ b/custom_components/audiconnect/__init__.py
@@ -3,12 +3,13 @@
 from __future__ import annotations
 
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any
 
 import homeassistant.helpers.config_validation as cv
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import HomeAssistant, ServiceCall
+from homeassistant.exceptions import ConfigEntryAuthFailed
 from homeassistant.helpers import device_registry as dr, entity_registry as er
 from homeassistant.helpers.device_registry import DeviceEntry
 
@@ -33,7 +34,10 @@ from .audi_account import (
 from .const import (
     CONF_API_LEVEL,
     CONF_DEVICE_ID,
+    CONF_PASSWORD,
+    CONF_REFRESH_TOKEN,
     CONF_SCAN_INITIAL,
+    CONF_USERNAME,
     DEFAULT_API_LEVEL,
     DOMAIN,
     PLATFORMS,
@@ -88,6 +92,17 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry) ->
             config_entry, version=2, data=new_data, options=new_options
         )
 
+    if config_entry.version == 2:
+        # v2 -> v3:
+        # Authentication moved from username/password (retired by Audi's move to
+        # Play Integrity attestation on the token exchange) to the OAuth Device
+        # Authorization Grant. Drop the stored credentials; the entry has no refresh
+        # token yet, so setup will start a reauthentication (device-code login).
+        new_data = {**config_entry.data}
+        new_data.pop(CONF_PASSWORD, None)
+        new_data.pop(CONF_USERNAME, None)
+        hass.config_entries.async_update_entry(config_entry, version=3, data=new_data)
+
     _LOGGER.info("Migration to version %s successful", config_entry.version)
     return True
 
@@ -98,6 +113,7 @@ class AudiRuntimeData:
 
     account: AudiAccount
     coordinator: AudiDataUpdateCoordinator
+    options_snapshot: dict[str, Any] = field(default_factory=dict)
 
 
 def _resolve_device_to_vin(hass: HomeAssistant, device_id: str) -> str | None:
@@ -157,6 +173,12 @@ def _get_all_coordinators(hass: HomeAssistant) -> list[AudiDataUpdateCoordinator
 async def _async_update_listener(
     hass: HomeAssistant, config_entry: ConfigEntry
 ) -> None:
+    runtime_data: AudiRuntimeData | None = getattr(config_entry, "runtime_data", None)
+    if runtime_data is not None and runtime_data.options_snapshot == dict(
+        config_entry.options
+    ):
+        # Only entry data changed (e.g. a rotated refresh token); no reload needed.
+        return
     await hass.config_entries.async_reload(config_entry.entry_id)
 
 
@@ -308,6 +330,12 @@ def _async_cleanup_orphaned_devices(
 
 async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry) -> bool:
     """Set up Audi Connect from a config entry."""
+    if not config_entry.data.get(CONF_REFRESH_TOKEN):
+        # No stored authorization (fresh v2->v3 migration): prompt device-code login.
+        raise ConfigEntryAuthFailed(
+            "Audi Connect needs to sign in again using device-code login"
+        )
+
     account = AudiAccount(hass, config_entry)
     coordinator = AudiDataUpdateCoordinator.from_entry(hass, account, config_entry)
 
@@ -316,7 +344,9 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry) -> b
 
     account.set_refresh_callback(_request_refresh)
     config_entry.runtime_data = AudiRuntimeData(
-        account=account, coordinator=coordinator
+        account=account,
+        coordinator=coordinator,
+        options_snapshot=dict(config_entry.options),
     )
 
     config_entry.async_on_unload(
--- a/custom_components/audiconnect/audi_account.py
+++ b/custom_components/audiconnect/audi_account.py
@@ -9,9 +9,14 @@ import voluptuous as vol
 
 import homeassistant.helpers.config_validation as cv
 from homeassistant.core import HomeAssistant, ServiceCall
+from homeassistant.exceptions import ConfigEntryAuthFailed
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
 
-from .audi_connect_account import AudiConnectAccount, AudiConnectObserver
+from .audi_connect_account import (
+    AudiAuthError,
+    AudiConnectAccount,
+    AudiConnectObserver,
+)
 from .audi_models import VehicleData
 from .const import (
     CONF_ACTION,
@@ -33,8 +38,7 @@ from .const import (
     CONF_UPDATE_SLEEP,
     CONF_SPIN,
     CONF_TARGET_SOC,
-    CONF_PASSWORD,
-    CONF_USERNAME,
+    CONF_REFRESH_TOKEN,
     DEFAULT_API_LEVEL,
     DOMAIN,
     REFRESH_VEHICLE_DATA_COMPLETED_EVENT,
@@ -118,17 +122,26 @@ class AudiAccount(AudiConnectObserver):
 
         self.connection = AudiConnectAccount(
             session=session,
-            username=self.config_entry.data.get(CONF_USERNAME),
-            password=self.config_entry.data.get(CONF_PASSWORD),
             country=self.config_entry.data.get(CONF_REGION),
             spin=self.config_entry.data.get(CONF_SPIN),
             api_level=self.config_entry.data.get(CONF_API_LEVEL, DEFAULT_API_LEVEL),
             excluded_vins=excluded_vins,
+            refresh_token=self.config_entry.data.get(CONF_REFRESH_TOKEN),
         )
         self.connection.add_observer(self)
+        self.connection.set_refresh_token_listener(self._persist_refresh_token)
 
     def set_refresh_callback(self, callback: Callable[[], Any]) -> None:
         self._refresh_callback = callback
+
+    def _persist_refresh_token(self, refresh_token: str) -> None:
+        """Persist a rotated IDK refresh token back to the config entry."""
+        if self.config_entry.data.get(CONF_REFRESH_TOKEN) == refresh_token:
+            return
+        self.hass.config_entries.async_update_entry(
+            self.config_entry,
+            data={**self.config_entry.data, CONF_REFRESH_TOKEN: refresh_token},
+        )
 
     def _build_vehicle_data(self, vehicle: Any) -> VehicleData:
         cfg_vehicle = VehicleData(self.config_entry)
@@ -138,7 +151,11 @@ class AudiAccount(AudiConnectObserver):
     async def async_refresh_data(self) -> list[VehicleData]:
         """Refresh cloud data and update discovered vehicles."""
         _LOGGER.debug("Starting refresh cloud data...")
-        if not await self.connection.update(None):
+        try:
+            updated = await self.connection.update(None)
+        except AudiAuthError as err:
+            raise ConfigEntryAuthFailed(str(err)) from err
+        if not updated:
             _LOGGER.warning("Failed refresh cloud data")
             raise RuntimeError("Failed refresh cloud data")
 
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -13,7 +13,7 @@ from typing import Any
 from aiohttp import ClientResponseError, ClientSession
 
 from .audi_api import AudiAPI
-from .audi_services import AudiService
+from .audi_services import AudiAuthError, AudiService
 from .util import get_attr, log_exception, parse_datetime, parse_float, parse_int
 
 _LOGGER = logging.getLogger(__name__)
@@ -27,10 +27,6 @@ ACTION_CHARGER = "charger"
 ACTION_WINDOW_HEATING = "window_heating"
 ACTION_PRE_HEATER = "pre_heater"
 ACTION_ENGINE = "engine"
-
-
-class AudiAuthError(Exception):
-    """Raised when the stored authorization is missing or no longer valid."""
 
 
 class AudiConnectObserver(ABC):
@@ -144,6 +140,10 @@ class AudiConnectAccount:
                     self._on_refresh_token_update(new_refresh_token)
             _LOGGER.debug("LOGIN: Audi session established")
             return True
+        except AudiAuthError:
+            # Token rejected (expired/revoked): propagate so Home Assistant reauth
+            # is triggered instead of retrying a credential that can never work.
+            raise
         except Exception as exception:
             if logError is True:
                 _LOGGER.error(
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -6,6 +6,7 @@ import re
 import time
 from abc import ABC, abstractmethod
 from asyncio import TimeoutError
+from collections.abc import Callable
 from datetime import datetime, timedelta, timezone
 from typing import Any
 
@@ -28,6 +29,10 @@ ACTION_PRE_HEATER = "pre_heater"
 ACTION_ENGINE = "engine"
 
 
+class AudiAuthError(Exception):
+    """Raised when the stored authorization is missing or no longer valid."""
+
+
 class AudiConnectObserver(ABC):
     @abstractmethod
     async def handle_notification(self, vin: str, action: str) -> None:
@@ -40,18 +45,16 @@ class AudiConnectAccount:
     def __init__(
         self,
         session: ClientSession,
-        username: str,
-        password: str,
         country: str,
         spin: str | None,
         api_level: int,
         excluded_vins: list[str] | None = None,
+        refresh_token: str | None = None,
     ) -> None:
         self._api = AudiAPI(session)
         self._audi_service = AudiService(self._api, country, spin, api_level)
 
-        self._username = username
-        self._password = password
+        self._refresh_token = refresh_token
         self._loggedin = False
         self._support_vehicle_refresh = True
         self._logintime: float = 0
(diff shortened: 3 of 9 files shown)


antoinevalentinHA:
Independent verification: **works end to end** on a different region and vehicle to yours.

**Setup:** Audi A3 Sportback e-tron (PHEV), region **FR** (EMEA), account that was previously failing with
`invalid assertion headers` (I posted the header probe table in #772).

I drove your branch's `AudiService` directly, outside Home Assistant, to isolate the auth
chain from everything else. All five steps pass:

1. `request_device_code()` — accepted with **no attestation header**.
   Endpoint used: `https://identity.vwgroup.io/oidc/v1/device_authorization`
   (the hardcoded fallback — see note below). `interval: 1`, `expires_in: 300`.
2. Browser approval at `https://identity.vwgroup.io/oidc/device/audi?user_code=...` — the
   normal Audi device page.
3. `poll_device_token()` → `ok`. Bearer token issued (`access_token` len 1200,
   `refresh_token` len 1251, `expires_in` 3599).
4. **Vehicle API accepts the token** — `get_vehicle_information()` returns the vehicle.
5. `login_with_refresh_token()` → succeeds, and the refresh token **rotates**
   (new token differs). Good to see the PR persists the rotated value back to the entry;
   that rotation is not optional to handle.

So the device-code grant is confirmed live for Audi in FR, not just DE/UK.

**One note on endpoint discovery.** The CARIAD OIDC discovery document at
`https://emea.bff.cariad.digital/auth/v1/idk/oidc/openid-configuration` does **not**
advertise a `device_authorization_endpoint` key, so your `_discover_endpoints()` always
falls through to the hardcoded `identity.vwgroup.io` URL. It works, but it means the
discovery branch is currently dead code and the flow depends on a constant. Worth a
comment in the code at least, since a future CARIAD change there would be silent.

Interestingly the discovery doc *does* list the grant:

```
"grant_types_supported": ["authorization_code","refresh_token","client_credentials",
                          "password","urn:ietf:params:oauth:grant-type:device_code",
                          "urn:ietf:params:oauth:grant-type:token-exchange"]
```

which is a nice confirmation the device grant is officially supported rather than an
accident.

**Still open:** I have only verified a single refresh cycle. Your ask for multi-day refresh
confirmation is the right one — I will run it in my HA and report back if the refresh
survives several days.

Nice piece of work. The device grant sidesteps attestation precisely because RFC 8628
targets input-constrained devices that cannot attest, so I would expect this to be durable
unless Audi deliberately restricts the grant for this client.

Kolbi:
I hopefully will have time tonight or tomorrow to check this PR and maybe merge it as fast as possible and push new version.

@coreywillwhat if you have time feel free to check.

mm98:
I've installed this PR. The `Audi Connect` devices are working, but I'm seeing this in the logs. Isn't really breaking anything for me. I just thought i would report it.

```python
2026-07-15 18:27:13.535 ERROR (MainThread) [custom_components.audiconnect.audi_connect_account] TRIP DATA: An unexpected error occurred while updating trip data for VIN: *************1567. Error: 'access_token'
2026-07-15 18:27:13.536 ERROR (MainThread) [custom_components.audiconnect.audi_connect_account] TRIP DATA: An unexpected error occurred while updating trip data for VIN: *************1567. Error: 'access_token'
2026-07-15 18:27:13.809 ERROR (MainThread) [custom_components.audiconnect.audi_connect_account] An unexpected error occurred while updating climater for VIN: *************1567. Error: can only concatenate str (not "NoneType") to str
2026-07-15 18:27:13.809 ERROR (MainThread) [custom_components.audiconnect.audi_connect_account] Unable to obtain the vehicle preheater state for *************1567: can only concatenate str (not "NoneType") to str
  File "/config/custom_components/audiconnect/audi_connect_account.py", line 1003, in update_vehicle_preheater
    result = await self._audi_service.get_preheater(self._vehicle.vin)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/audiconnect/audi_services.py", line 121, in get_preheater
    return await self._api.get(
           ^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
    )
    ^
  File "/config/custom_components/audiconnect/audi_api.py", line 184, in get
    full_headers = self.__get_headers()
  File "/config/custom_components/audiconnect/audi_api.py", line 247, in __get_headers
    data["Authorization"] = "Bearer " + self.__token.get("access_token")
                            ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TypeError: can only concatenate str (not "NoneType") to str
```

_I dont have preheating in the car. Audi A5 Sportback (2021) ICE_

antoinevalentinHA:
@Kolbi @jonny190 — heads-up before this merges and ships, I found a real bug running it in my HA.

**The device-code flow breaks the entire MBB path.** `DEVICE_CODE_SCOPE` omits `mbb`, which the old password flow requested. That scope is what puts the MBB audiences into the id_token that `_finalize_session()` feeds to mbboauth:

```
POST mbboauth .../mobile/oauth2/v1/token
-> 400 {"error":"invalid_grant","error_description":
        "Audiences ([09b6cbec-...@apps_vw-dilab_com]) don't match issuer (VWGMBB01DELIV1)."}
```

Adding `mbb` puts `VWGMBB01DELIV1` in the id_token audience and mbboauth returns 200. One word.

The response body is currently stored unchecked, so `vwToken` silently becomes an error dict and only blows up much later, far from the cause:

```
TRIP DATA: ... Error: 'access_token'                                            (x20)
... updating climater ... Error: can only concatenate str (not "NoneType") to str  (x10)
```

Affected: trip data, climater, charger, timers, `get_operations_list`, `_get_security_token` — **all vehicle commands**, not just telemetry. The CARIAD BFF endpoints use the IDK bearer token and are genuinely fine, which is why verification against them passes; only MBB is hit, silently.

Fix + explanation: jonny190/audi_connect_ha#2. Verified live on an e-tron/FR — with `mbb`, mbboauth returns 200 and vwToken carries a real access_token, and the BFF path is unaffected (device code, vehicle API, refresh with rotation all still pass).

Also worth knowing before you cut a release: the translations. #777 updates `en.json` only, and HA does not fall back to English for custom integration translations, so the device step renders **blank** in the 7 other shipped languages — setup is impossible for those users. I hit that in French myself. #778 fixes French (merge after this one); the other six need native speakers.

coreywillwhat:
I won't have time to check this today. But as far as I'm aware the US is not broken and has survived restarts. How are different regions handled in this PR?

coreywillwhat:
Thinking we push this through beta/pre-release and see how it works for everyone in all regions since this is such a large change.

Kolbi:
> I won't have time to check this today. But as far as I'm aware the US is not broken and has survived restarts. How are different regions handled in this PR?

If US is still working, I guess this PR would break it as I can't see a switch....

antoinevalentinHA:
Some data on the region question, and a map of what's still open, in case it helps you plan the beta.

**How regions are handled (answering @coreywillwhat).** The PR already routes by region: `__get_cariad_url()` picks `emea` vs `na` on `country == "US"`, and the token endpoint is taken from *that* region's discovery document — so a US entry exchanges tokens on `na.bff.cariad.digital`, never EMEA. The device-authorization step (where the user approves the code) uses the global `identity.vwgroup.io`, which is the same for every region. So the only region-specific parts are token exchange and vehicle data, and both already follow the entry's country.

**Does device-code work in the US?** The US backend advertises it: `https://na.bff.cariad.digital/auth/v1/idk/oidc/openid-configuration` lists `urn:ietf:params:oauth:grant-type:device_code` in `grant_types_supported`, identical to EMEA. That's a strong sign it'll work — but I can only verify EMEA end to end (I'm in FR, no US account). **This is exactly what the beta is for**, so +1 to the pre-release plan: it confirms the US path on real US accounts before it reaches everyone.

**Two fixes worth folding into the branch before that beta ships:**

1. **The `mbb` scope** — @jonny190 already landed this (`e6dc382`). Without it, login *succeeds* but every legacy MBB call (lock/unlock, trip data, climater, S-PIN) fails, because the device-code token was missing the `mbb` audience. Several people hit it: @mm98's A5, my A3. ✅ done.

2. **jonny190/audi_connect_ha#2 — fail loudly on a rejected MBB token.** Right now that failure is stored silently and only surfaces hours later as a cryptic `KeyError('access_token')` in unrelated code. This PR raises a proper auth error instead. It matters for the beta specifically: it's also the **automatic migration** for anyone already signed in. Their stored token predates the `mbb` scope and keeps refreshing happily, so nothing ever re-prompts them — they'd stay silently half-broken after updating. The guard turns that into a normal "re-authenticate" prompt that reissues the token *with* `mbb`. Without it you'd need a config-version bump to force reauth.

3. **#778 — translations.** The device step is only in `en.json`; HA doesn't fall back to English for custom integrations, so it renders **blank** in the 7 other shipped languages and setup can't be completed in them. #778 fills all seven (native FR, the rest carefully done and open to native corrections).

Happy to help test whatever you need on EMEA.

coreywillwhat:
Thank you for this! This was pushed as a pre-release. If there are additional PRs to resolve bugs within this PR, please create PRs to merge with the `beta` branch and we can push more pre-releases until this is stable.

coreywillwhat:
I will test for the US when I have some time.

MonacoLM:
What's now the step by step process to get a new car setup?
When trying to add the integration, I get the previous config screen but with without user and pasword fields. By clicking on OK, I get an "Audi Connect" popup with a "submit" button.  That button isn't doing anything.  If I close the pop-up, the integration has not been added.
What am I missing?

tinkano:
> What's now the step by step process to get a new car setup? When trying to add the integration, I get the previous config screen but with without user and pasword fields. By clicking on OK, I get an "Audi Connect" popup with a "submit" button. That button isn't doing anything. If I close the pop-up, the integration has not been added. What am I missing?

Make a reboot not restart system

DJm00n:
I can confirm that [v2.2.1b1](https://github.com/audiconnect/audi_connect_ha/releases/tag/v2.2.1b1) works on my side:
image

Thank you @jonny190 and @coreywillwhat!

its-me-prash:
@coreywillwhat I opened #785 with a few small fixes on top of this. It covers the empty reconfigure popup some people ran into in #775, the raw reconfigure_successful key that shows up after saving the reconfigure dialog and the login texts telling people to press Submit while the button is actually called OK now.

And since quite a few people over in #775 confirmed b1 is working fine for them now, maybe worth considering to promote this to a normal release once the small stuff is in. Everyone on the stable channel is still sitting on 2.2.0 with broken login and most of them will never see a pre release.

coreywillwhat:
Thank you for these! I haven't seen anyone in the US report they tested or are using the b1 release, so we won't push to stable until at least I have tested. I'll review and merge new PRs and push a new prerelease over the weekend and I'll try to find some time to test myself.

MonacoLM:
> > What's now the step by step process to get a new car setup? When trying to add the integration, I get the previous config screen but with without user and pasword fields. By clicking on OK, I get an "Audi Connect" popup with a "submit" button. That button isn't doing anything. If I close the pop-up, the integration has not been added. What am I missing?
> 
> Make a reboot not restart system

Thanks!  [v2.2.1b1](https://github.com/audiconnect/audi_connect_ha/releases/tag/v2.2.1b1) working for me now.

coreywillwhat:
Unfortunately, I could not get this to work in the US. FYI to anyone else -- Downgrading to v2.2.0 requires removal of the integration and then readding/reconfiguring.

There are two reports (at the moment) of v2.2.0 beginning to work again for a user in EU: https://github.com/audiconnect/audi_connect_ha/issues/772#issuecomment-5024349817

is someone able to downgrade and test? @Kolbi thoughts?

Kolbi:
> Unfortunately, I could not get this to work in the US. FYI to anyone else -- Downgrading to v2.2.0 requires removal of the integration and then readding/reconfiguring.
> 
> There are two reports (at the moment) of v2.2.0 beginning to work again for a user in EU: [#772 (comment)](https://github.com/audiconnect/audi_connect_ha/issues/772#issuecomment-5024349817)
> 
> is someone able to downgrade and test? @Kolbi thoughts?

I think before Thursday I have no chance to test the rollback to v2.2.0.

antoinevalentinHA:
For what it's worth — I'm a user here, not a maintainer, so please take this as data points rather than a verdict. I put together a timeline that might help before the rollback decision.

The "v2.2.0 works again" reports look to me like VW toggling the Play Integrity enforcement on and off, rather than a real fix:

- ~13 Jul: password flow breaks
- 20 Jul: several EU users report v2.2.0 works again, no reconfig (BE, DE, NL)
- 22 Jul: it breaks again — @ekirchma and @mkowa42 (EU/DE): *"worked for two days, login error reoccurred today"*

If that reading is right, v2.2.0 isn't fixed — it's just currently un-enforced for some accounts and already re-enforced for others. The appeal of the device-code flow is that it sidesteps attestation entirely, so it holds whichever way VW toggles; a rollback would re-break the moment enforcement returns, which seems to have happened on the 22nd.

The US failure looks like a genuinely separate problem (something in the US-side token exchange / market config, not the auth *method*), and a rollback wouldn't solve it either, since v2.2.0 depends on attestation too. If someone with a US account can share the debug-log error from the device-code step, I'm glad to help dig into it — though I can't promise much without a US account to reproduce on.

(Full disclosure: I work through the technical side of this with an AI assistant, so please weigh the analysis on its own merits rather than on my say-so — I'm just relaying what holds up.)

coreywillwhat:
Thanks @antoinevalentinHA. I like the solution presented in the beta, but also, the US v2.2.0 is still functioning and has not broken since these issues began a few weeks ago. Ideally, we continue to use the existing structure for the US from v2.2.0, and fold in the solution for the EU so its region based. If the US eventually breaks for the same reasons, we could then work to get the US Working. Let me know your thoughts.

antoinevalentinHA:
That makes sense to me — if the US path still works, breaking it to force a flow you couldn't get working there would be the worse trade. Region-based sounds like the right call.

One consequence I should flag on my own PR: **#781 should probably be closed.** It removes the HTML-scraping helpers (`get_hidden_html_input_form_data`, `get_post_url`, `_to_absolute`) and the `beautifulsoup4` requirement — which are only dead code if the password flow is gone for good. If the US keeps the v2.2.0 structure, they're live again. Happy to close it myself, just say the word. (Your instinct to hold it was right.)

For the region switch, there may already be a natural hook: the code branches on `self._country == "US"` in a few places (the CARIAD host `emea`/`na`, and some vehicle-action paths), so the auth method could key off the same region value from the config entry rather than needing new plumbing.

The one thing I'd gently keep on the radar: if VW enforces attestation in the US the way it did in the EU, that path breaks with no fallback ready. It might be worth capturing a debug log of the US device-code failure while there's no time pressure, so the diagnosis exists before it's urgent. Happy to help look if someone can share one — though without a US account I can't reproduce.

coreywillwhat:
Time is my biggest issue these days.

Would you or someone else be open to trying to create a new PR which maintains the US and other region methods, with the specific (beta) solution for the EU region?

I can grab the logs for the US at some point based on the current beta, but i would expect if we have both solutions working in the same build, we could easily switch a couple lines of code to incorporate the US into the new method once its ironed out.

antoinevalentinHA:
Happy to take this on.

To set expectations up front: I can implement and test the **EU** path (device-code) on my own account, but I have no US account — so the restored password path will need your validation before merge. I'll keep the two paths cleanly separated so that's straightforward to test.

Plan, unless you'd rather it differently:

- Restore the v2.2.0 password login (`login_request` plus the HTML-form helpers and `beautifulsoup4`) alongside the device-code flow. It's all still in git history, so this is a restore rather than a rewrite.
- Switch on the region already stored in the config entry: **Europe → device-code**, **USA / Canada / China → password**. Say the word if you'd rather Canada/China went the other way — as far as I can tell nobody has tested those against the beta.
- Config flow branches on region: username/password step for the legacy regions, device-code step for Europe, and the same split for reauth.

One thing worth flagging that isn't obvious: **the current v2→v3 migration drops the stored credentials unconditionally** — `__init__.py` pops `CONF_USERNAME`/`CONF_PASSWORD` for every entry. As it stands, a US user upgrading from v2.2.0 would have their credentials wiped and then be pushed into device-code, which is exactly the breakage we're trying to avoid. I'll make that migration region-aware as part of the PR.

I'll open it against `beta`.

PR #781, opened 2026-07-16
chore: drop dead password-flow code and the beautifulsoup4 requirement

antoinevalentinHA:
Follow-up cleanup on the `beta` branch after the device-code migration (#777).

## What

The username/password login is gone, but its HTML-scraping helpers were left behind. These now have **zero callers**:
- `AudiService.get_hidden_html_input_form_data`
- `AudiService.get_post_url`
- `_to_absolute` (used only by `get_post_url`)

`BeautifulSoup` and `urlparse` are imported solely for those, and `beautifulsoup4` is the manifest's only requirement — also unused now.

## Changes
- Remove the three dead functions and the `bs4` / `urlparse` imports.
- Drop `beautifulsoup4` from `requirements` (now `[]`).

## Why it's safe
Pure dead-code removal — `grep` confirms no remaining references to any of the removed names anywhere in the integration. No behaviour change.

**Bonus:** the requirement was unpinned, which fails hassfest's requirements validation (same family as the `version` issue noted on #780). Removing it clears that too.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -8,9 +8,7 @@ import logging
 from datetime import datetime, timedelta, timezone
 from hashlib import sha512
 from typing import Any
-from urllib.parse import urlencode, urlparse
-
-from bs4 import BeautifulSoup
+from urllib.parse import urlencode
 
 from .audi_api import AudiAPI
 from .audi_models import TripDataResponse, VehicleDataResponse, VehiclesResponse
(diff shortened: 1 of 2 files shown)


coreywillwhat:
Pending testing in the US. If beta doesn't work in the US, we'll need to reconsider the config_flow

coreywillwhat:
As discussed, these are still required for regions outside the EU. Regardless, i appreciate the cleanup effort!

PR #791, opened 2026-07-22
feat: region-based authentication (device-code for Europe, password elsewhere)

antoinevalentinHA:
Implements the region-based approach discussed in #777: keep the working password login for the regions Audi hasn't touched, and use the Device Authorization Grant where attestation forces it.

## How the choice is made

One decision point, in `const.py`:

```python
DEVICE_CODE_REGIONS: set[str] = {REGION_EUROPE}

def uses_device_code(region: str | None) -> bool:
    return (region or REGION_EUROPE) in DEVICE_CODE_REGIONS
```

When attestation reaches another region, add it to that set — the config flow, the setup guard and the migration all read from this one helper, so nothing else has to change. That's the "switch a couple lines of code" you were after.

Currently: **Europe → device-code**, **USA / Canada / China → password**. Say the word if you'd rather Canada/China went the other way; I grouped them with the US because nothing suggests attestation is enforced there, but nobody has tested them.

## What changed

- **`audi_services.py`** — `login()` / `login_request()` restored. Rather than pasting the old method back, it is rebuilt on the helpers #777 extracted: `_discover_endpoints()` for the prologue and `_finalize_session()` for the AZS/mbboauth tail, so both flows share that code. The token exchange now raises `AudiAuthError` carrying the backend's own error instead of failing later with a `KeyError` — in Europe that's exactly where `invalid assertion headers` surfaces.
- **`audi_connect_account.py`** — takes either a refresh token or username/password; `try_login()` dispatches.
- **`config_flow.py`** — branches after the region is picked: credentials step for legacy regions, device-code step for Europe, same split on reauth.
- **Translations** — `credentials` / `reauth_credentials` steps and `invalid_credentials` added to all 8 shipped languages, reusing the labels that already existed before #777 (no re-translation).

## Migration bug fixed along the way

The v2→v3 migration dropped `CONF_USERNAME` / `CONF_PASSWORD` **unconditionally**. As it stood, a US user upgrading from v2.2.0 would have had working credentials wiped and then been pushed into device-code — the exact breakage this PR exists to prevent. It's now region-aware, and `async_setup_entry` raises a clear reauth prompt for whichever credential type the region expects instead of failing obscurely.

## Testing

- **Europe: verified end to end** just now on my own account against this branch — device code → token → vehicle API → refresh with rotation, all pass. No regression from the refactor.
- **US / Canada / China: not tested.** I have no account outside the EU. The password path is restored from the pre-#777 code, so it should behave exactly as v2.2.0 did, but it needs your confirmation before merge.

## Note on #781

This supersedes #781 — `beautifulsoup4` and the HTML-form helpers are load-bearing again for the password path, so that PR should be closed. I'll close it once you confirm this direction.

### DIFF
--- a/custom_components/audiconnect/__init__.py
+++ b/custom_components/audiconnect/__init__.py
@@ -36,11 +36,13 @@ from .const import (
     CONF_DEVICE_ID,
     CONF_PASSWORD,
     CONF_REFRESH_TOKEN,
+    CONF_REGION,
     CONF_SCAN_INITIAL,
     CONF_USERNAME,
     DEFAULT_API_LEVEL,
     DOMAIN,
     PLATFORMS,
+    uses_device_code,
 )
 from .coordinator import AudiDataUpdateCoordinator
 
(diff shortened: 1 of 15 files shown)


coreywillwhat:
I will test this PR in pre-release and report back.

coreywillwhat:
Tested the latest pre-release [v2.2.1b3](https://github.com/audiconnect/audi_connect_ha/releases/tag/v2.2.1b3) in the US.
Updating from v2.2.0: No issues. Everything continued to work as expected.
I then removed the integration, restarted, and readded the integration to simulate a new install, and the config flow works as expected with no issues. Lets give the pre-release a few days and ensure there are no issues from other users, then I think we can push to production/latest release. 

@Kolbi If you could test and give any input I'd appreciate it. I know you're busy until thursday or so. No pressure or rush, but always appreciate your input.

coreywillwhat:
Oh and thank you for your work on this @antoinevalentinHA !! Really appreciate the collaboration and quick effort you've put in!

Kolbi:
> @Kolbi If you could test and give any input I'd appreciate it. I know you're busy until thursday or so. No pressure or rush, but always appreciate your input.

Latest [v2.2.1b4](https://github.com/audiconnect/audi_connect_ha/tree/v2.2.1b4) is installed and working for me.

Issue #802, opened 2026-07-27
"Invalid credentials" error during login / setup despite correct credentials


hikai5:
When trying to configure / set up the Audi Connect integration, I receive the error message "Ungültige Anmeldeinformationen" (Invalid credentials). However, I am able to successfully log in to the myAudi website and the official myAudi mobile app using the exact same credentials (username/email and password).

I have already waited over 24 hours since my last setup attempt to rule out temporary IP or account rate-limiting, but the issue persists.

To Reproduce
Go to Settings -> Devices & Services -> Add Integration.
Select Audi Connect.
Enter username (email) and password.
Click on Submit / Next.

See error: "Ungültige Anmeldeinformationen" (Invalid credentials).

Expected behavior
The integration should authenticate successfully and discover/add the vehicle entities.

Environment / Versions
Home Assistant Core Version: 2026.7.4
Audi Connect Integration Version: 2.2.0
Vehicle Model: Audi Q5 FY 2019
Country / Region: Germany (EU)

Additional context
Logged in successfully on myAudi app and web portal using identical credentials.
Waited >24 hours to rule out temporary rate limiting / lockouts.
Special characters in password: NO

<img width="554" height="788" alt="Image" src="https://github.com/user-attachments/assets/08219a18-56ea-4ee5-a01a-fb4bdd39962d" />

hikai5:
sorry, didn't see https://github.com/audiconnect/audi_connect_ha/issues/775

Issue #805, opened 2026-07-28
Failed refresh cloud data


Audi678:
**Describe the bug**
It works fine, but now in think after updating HA core the respons is "Failed refresh cloud data"

**To Reproduce**

**Expected behavior**

**Screenshots**

Audi Connect
Versie 2.2.0
Instellen mislukt, wordt opnieuw geprobeerd: Failed refresh cloud data
A3 etron ambition pro-line
A3 Sportback Ambition TFSI S tronic (2015) • Audi A3 Etron • 49 entiteiten

Hubs
Geen items
**Logfile**

 LOGIN: Login to Audi service failed, retrying in 10 seconds 
 15:21:02 – (WAARSCHUWING) Audi Connect (aangepaste integratie) - bericht kwam voor het eerst om 14:57:18 en verschijnt 17 keer
 Failed refresh cloud data 
 15:11:00 – (WAARSCHUWING) Audi Connect (aangepaste integratie) - bericht kwam voor het eerst om 14:57:41 en verschijnt 8 keer
 LOGIN: Failed to log in to the Audi service: 'access_token'. If this error persists, open the myAudi app or log in via a web browser and accept any pending terms or consent prompts, then restart the integration. 
 15:11:00 – (FOUT) Audi Connect (aangepaste integratie) - bericht kwam voor het eerst om 14:57:41 en verschijnt 8 keer

How to enable audiconnect debugging?
Settings > Integrations > Audi Connect > Enable Debug Logging
Run service refresh_cloud_data
Disable Debug Logging

**Your Vehicle Details**
Model: A3 Etron
Year: 2015
Type Hybrid
Region EU

alantaTAlbers:
Using v2.2.0 and have still (or perhaps again) the same problem: Einrichtungsfehler, wird erneut versucht: Failed refresh cloud data

No Message Popup in the Audi App which might prevent successfull login.


> 2026-07-30 14:37:15.156 DEBUG (MainThread) [custom_components.audiconnect.audi_connect_account] LOGIN: Requesting login to Audi service...
> 2026-07-30 14:37:15.156 DEBUG (MainThread) [custom_components.audiconnect.audi_services] LOGIN: Starting login to Audi service...
[...]

PR #806, opened 2026-07-28
fix: use myAudi device-code flow to bypass broken token exchange

### LINKED ISSUE #772: Login fails with KeyError('access_token') — CARIAD token endpoint returns 400 "invalid assertion headers"; hybrid flow blocked by "invalid client"

Rothgar666:
## Problem

Since around 2026-05, CARIAD binds the `authorization_code` exchange to a Play Integrity / `x-assertion` header that only the genuine myAudi app can produce. Exchanging the code at the IDK token endpoint now returns `HTTP 400 {"error":"invalid assertion headers"}`, which surfaces in Home Assistant as "Invalid credentials" during setup even though the credentials are correct.

Fixes #772, #775, #802.

## Fix

Replace the authorization-code login with the OAuth 2.0 **device-code flow**:

- `device_authorization` -> poll `identity.vwgroup.io/oidc/v1/token`
- The user confirmation is driven server-side with username/password, so login stays headless (no user interaction).
- Tokens are issued directly by `identity.vwgroup.io` and accepted by the BFF **without** any assertion header.
- The flow yields a `refresh_token`, so token refresh now runs against the BFF without an assertion header or `client_secret`.
- The legacy `vwToken` is re-derived from the refreshed `id_token`, keeping all existing read/PIN paths working.

## Compatibility

No breaking changes: `login_request` sets the exact same token attributes as before (`vwToken`, `audiToken`, `mbboauthToken`, `xclientId`, `_bearer_token_json`), no public method signatures changed, and the S-PIN paths (lock/unlock, engine, pre-heater) use the same tokens as before.

## Testing

Verified end-to-end against a live myAudi account: login succeeds and the BFF `/vehicles` endpoint returns the vehicle list.

### DIFF
--- a/custom_components/audiconnect/audi_api.py
+++ b/custom_components/audiconnect/audi_api.py
@@ -45,6 +45,15 @@ class AudiAPI:
         if DEBUG_VERBOSE:
             _LOGGER.debug("[set_xclient_id] X-Client-ID set: %s", xclientid)
 
+    def clear_cookies_for_domains(self, domains: tuple[str, ...]) -> None:
+        # Best-effort: never let a cookie jar quirk block login.
+        try:
+            jar = self._session.cookie_jar
+            for domain in domains:
+                jar.clear_domain(domain)
+        except Exception:
+            _LOGGER.debug("Could not clear auth cookies before login", exc_info=True)
+
     async def request(
         self,
         method: str,
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -1,17 +1,14 @@
 from __future__ import annotations
 
 import asyncio
-import base64
 import hmac
 import json
 import logging
-import os
 import re
-import uuid
 from datetime import datetime, timedelta, timezone
-from hashlib import sha256, sha512
+from hashlib import sha512
 from typing import Any
-from urllib.parse import parse_qs, urlencode, urlparse
+from urllib.parse import urlencode, urlparse
 
 from bs4 import BeautifulSoup
 
@@ -23,6 +20,7 @@ from .util import get_attr, to_byte_array
 
 MAX_RESPONSE_ATTEMPTS = 10
 REQUEST_STATUS_SLEEP = 10
+MAX_LOGIN_REDIRECTS = 20
 
 SUCCEEDED = "succeeded"
 FAILED = "failed"
@@ -971,7 +969,7 @@ class AudiService:
                         )
                     )
 
-        raise Exception("Request {} timed out".format(request_id))
+        raise Exception(f"Request {request_id} timed out")
 
     async def check_request_succeeded(
         self, url: str, action: str, successCode: str, failedCode: str, path: str
@@ -994,7 +992,7 @@ class AudiService:
             if status == successCode:
                 return
 
-        raise Exception("Cannot {action}, operation timed out".format(action=action))
+        raise Exception(f"Cannot {action}, operation timed out")
 
     # TR/2022-12-20: New secret for X_QMAuth
     def _calculate_X_QMAuth(self) -> str:
@@ -1053,80 +1051,55 @@ class AudiService:
     # TR/2021-12-01: Refresh token before it expires
     # returns True when refresh was required and successful, otherwise False
     async def refresh_token_if_necessary(self, elapsed_sec: int) -> bool:
-        if self.mbboauthToken is None:
+        # The device-code flow issues an IDK bearer token with its own
+        # refresh_token. The BFF refreshes it without any assertion header or
+        # client_secret, so we drive the refresh off _bearer_token_json.
+        if self._bearer_token_json is None:
             return False
-        if "refresh_token" not in self.mbboauthToken:
+        if "refresh_token" not in self._bearer_token_json:
             return False
-        if "expires_in" not in self.mbboauthToken:
+        if "expires_in" not in self._bearer_token_json:
             return False
 
-        if (elapsed_sec + 5 * 60) < self.mbboauthToken["expires_in"]:
+        if (elapsed_sec + 5 * 60) < self._bearer_token_json["expires_in"]:
             # refresh not needed now
             return False
 
         try:
+            # Refresh the IDK bearer token directly against the Cariad BFF.
             headers = {
                 "Accept": "application/json",
                 "Accept-Charset": "utf-8",
                 "User-Agent": AudiAPI.HDR_USER_AGENT,
                 "Content-Type": "application/x-www-form-urlencoded",
-                "X-Client-ID": self.xclientId,
             }
-            mbboauth_refresh_data = {
-                "grant_type": "refresh_token",
-                "token": self.mbboauthToken["refresh_token"],
-                "scope": "sc2:fal",
-                # "vin": vin,  << App uses a dedicated VIN here, but it works without, don't know
-            }
-            encoded_mbboauth_refresh_data = urlencode(
-                mbboauth_refresh_data, encoding="utf-8"
+            bearer_refresh_data = urlencode(
+                {
+                    "client_id": self._client_id,
+                    "grant_type": "refresh_token",
+                    "refresh_token": self._bearer_token_json["refresh_token"],
+                },
+                encoding="utf-8",
             ).replace("+", "%20")
-            mbboauth_refresh_rsp, mbboauth_refresh_rsptxt = await self._api.request(
-                "POST",
-                self.mbbOAuthBaseURL + "/mobile/oauth2/v1/token",
-                encoded_mbboauth_refresh_data,
-                headers=headers,
-                allow_redirects=False,
-                rsp_wtxt=True,
-            )
-
-            # this code is the old "vwToken"
-            self.vwToken = json.loads(mbboauth_refresh_rsptxt)
-
-            # TR/2022-02-10: If a new refresh_token is provided, save it for further refreshes
-            if "refresh_token" in self.vwToken:
-                self.mbboauthToken["refresh_token"] = self.vwToken["refresh_token"]
-
-            # hdr
-            headers = {
-                "Accept": "application/json",
-                "Accept-Charset": "utf-8",
-                "X-QMAuth": self._calculate_X_QMAuth(),
-                "User-Agent": AudiAPI.HDR_USER_AGENT,
-                "Content-Type": "application/x-www-form-urlencoded",
-            }
-            # IDK token request data
-            tokenreq_data = {
-                "client_id": self._client_id,
-                "grant_type": "refresh_token",
-                "refresh_token": self._bearer_token_json.get("refresh_token"),
-                "response_type": "token id_token",
-            }
-            # IDK token request
-            encoded_tokenreq_data = urlencode(tokenreq_data, encoding="utf-8").replace(
-                "+", "%20"
-            )
             bearer_token_rsp, bearer_token_rsptxt = await self._api.request(
                 "POST",
                 self._tokenEndpoint,
-                encoded_tokenreq_data,
+                bearer_refresh_data,
                 headers=headers,
                 allow_redirects=False,
                 rsp_wtxt=True,
             )
-            self._bearer_token_json = json.loads(bearer_token_rsptxt)
+            refreshed = json.loads(bearer_token_rsptxt)
+            if "access_token" not in refreshed:
+                raise Exception("Bearer token refresh returned no access_token")
+            # The BFF may omit a fresh refresh_token; keep the previous one then.
+            if "refresh_token" not in refreshed and "refresh_token" in (
+                self._bearer_token_json or {}
+            ):
+                refreshed["refresh_token"] = self._bearer_token_json["refresh_token"]
+            self._bearer_token_json = refreshed
 
-            # AZS token
+            # AZS token (Audi-specific services)
             headers = {
                 "Accept": "application/json",
                 "Accept-Charset": "utf-8",
@@ -1149,8 +1122,36 @@ class AudiService:
                 allow_redirects=False,
                 rsp_wtxt=True,
             )
-            azs_token_json = json.loads(azs_token_rsptxt)
-            self.audiToken = azs_token_json
+            self.audiToken = json.loads(azs_token_rsptxt)
+
+            # Re-derive the legacy vwToken from the refreshed id_token.
+            if "id_token" in self._bearer_token_json and self.xclientId:
+                headers = {
+                    "Accept": "application/json",
+                    "Accept-Charset": "utf-8",
+                    "User-Agent": AudiAPI.HDR_USER_AGENT,
+                    "Content-Type": "application/x-www-form-urlencoded",
+                    "X-Client-ID": self.xclientId,
+                }
+                mbboauth_auth_data = urlencode(
+                    {
+                        "grant_type": "id_token",
+                        "token": self._bearer_token_json["id_token"],
+                        "scope": "sc2:fal",
+                    },
+                    encoding="utf-8",
+                ).replace("+", "%20")
+                _mbb_rsp, mbb_txt = await self._api.request(
+                    "POST",
+                    self.mbbOAuthBaseURL + "/mobile/oauth2/v1/token",
+                    mbboauth_auth_data,
+                    headers=headers,
+                    allow_redirects=False,
+                    rsp_wtxt=True,
+                )
+                mbb_json = json.loads(mbb_txt)
+                self.mbboauthToken = mbb_json
+                self.vwToken = mbb_json
 
             return True
 
@@ -1220,10 +1221,6 @@ class AudiService:
         openidcfg_json = await self._api.request("GET", openidcfg_url, None)
 
         # use dynamic config from openId config
-        authorization_endpoint = "https://identity.vwgroup.io/oidc/v1/authorize"
-        if "authorization_endpoint" in openidcfg_json:
-            authorization_endpoint = openidcfg_json["authorization_endpoint"]
-
         self._tokenEndpoint = self.__get_cariad_url("/auth/v1/idk/oidc/token")
 
         if "token_endpoint" in openidcfg_json:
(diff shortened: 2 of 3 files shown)


mstu01:
The device-code flow works — thanks for this. Tokens are issued by `identity.vwgroup.io` and the BFF accepts them without any assertion header, exactly as described.

One issue on a normal Home Assistant install, though: setup failed for me at the identifier step with

```
LOGIN: Failed to log in to the Audi service: Could not extract CSRF/HMAC/relayState from the password page. Audi may be showing a consent or terms-of-service prompt.
```

That message is a bit of a red herring — there was no pending consent on the account. The real cause is the **shared aiohttp session**.

The integration builds `AudiService` on `async_get_clientsession(hass)` (config_flow.py / audi_account.py), which is HA's long-lived, shared session. After one or more login attempts it holds an authenticated SSO cookie for `identity.vwgroup.io`. On the next `login_request`, the IdP sees an existing session and serves an interstitial/consent page **instead of** the password form, so `csrf_token` / `hmac` / `relayState` aren't present and extraction fails. It's fully reproducible and self-inflicted by the retry loop.

Evidence — same account, same client_id, minutes apart:

| | redirect chain after `verification_uri_complete` | `/login/identifier` response |
|---|---|---|
| fresh session (`requests` or a fresh `aiohttp.ClientSession`) | `302 → 200` | password page, 4782 B, markers present |
| HA shared session (with the leftover SSO cookie) | `302 → 302 → 302 → 200` | consent page, ~5886 B, markers absent |

A fresh session always reaches the password page; the shared one, once it has the cookie, never does. Because every `refresh cloud data` cycle re-authenticates on the same shared session, the first login of a run can succeed and set the cookie, and every subsequent re-login then fails — which is why it also shows up as a retry loop during setup.

Fix that worked here: clear the auth-domain cookies from the session jar at the start of `login_request`, so every login starts from a clean state:

```python
async def login_request(self, user: str, password: str):
    self._api.use_token(None)
    self._api.set_xclient_id(None)
    self.xclientId = None

    # The shared HA aiohttp session accumulates an authenticated SSO cookie
    # from earlier logins. On a re-login the IdP then serves a consent page
    # instead of the password page, which breaks the device-flow form parsing
    # (CSRF/HMAC/relayState missing). Start each login from a clean cookie state.
    try:
        jar = self._api._session.cookie_jar
        for _dom in ("vwgroup.io", "cariad.digital", "vwg-connect.com"):
            jar.clear_domain(_dom)
    except Exception:  # best effort - never block login on jar quirks
        _LOGGER.debug("Could not clear auth cookies before login", exc_info=True)

    # get markets
    ...
```

With this, setup goes straight to the password page, the device flow completes, and the vehicle entities populate against a live account. Happy to open a PR against your branch if that's easier than folding it in.

Kolbi:
Hi @Rothgar666 would you mind to check your PR against existing https://github.com/audiconnect/audi_connect_ha/tree/beta branch?

antoinevalentinHA:
For context that might save some duplicated effort: the `beta` branch already ships this device-code flow — `request_device_code()` / `poll_device_token()` in `audi_services.py`, hitting `identity.vwgroup.io/oidc/v1/device_authorization`, with an `async_step_device` config step. It's been live in the `2.2.1bx` pre-releases for ~3 weeks and the same root cause (Play Integrity attestation on the `authorization_code` exchange → `400 invalid assertion headers`) is what it was built to fix. On my own EU account it's been stable since.

So this is really the same fix landing on `master` — which matters, since everyone in #775 / #802 / #805 is on the stable line that doesn't have it yet. Might be simplest to reconcile against `beta` (as @Kolbi suggested) rather than carry a parallel implementation.

On @mstu01's CSRF/HMAC extraction error: worth checking whether `beta`'s device step hits the same wall, since it drives the confirmation server-side too — if beta already gets past it, the difference in how the password page is fetched there might be the missing piece.

Happy to help test on EU if useful.

Rothgar666:
Thanks for the very detailed write-up — this is really useful. 

To make sure I've understood it correctly: the device-code flow itself works fine, and the failure at the identifier step is not a pending consent but the shared aiohttp session. Since AudiService is built on async_get_clientsession(hass), the session keeps the authenticated SSO cookie from an earlier login. On the next login_request the IdP then sees an existing session and serves an interstitial/consent page instead of the password form, so csrf_token / hmac / relayState aren't there and the extraction fails.

Your side-by-side comparison makes that pretty conclusive — fresh session goes 302 → 200 and lands on the password page with the markers present, while the shared session takes the extra redirects and ends up on a larger consent page without them. It also explains the retry loop: the first login of a run succeeds and sets the cookie, and every re-login after that fails, including during setup and on each refresh of the cloud data.

The fix of clearing the cookies for the auth domains at the start of login_request looks reasonable to me, and keeping it best-effort so a jar quirk can never block the login seems right.

I'd like to go through all of this again this evening and reproduce it on my own install before I take it further. I'll report back after that.

mstu01:
Yes, you've got it exactly right — that's the whole chain.

Two things that should save you time reproducing it tonight:

- **Forcing it deterministically:** you don't need to wait for the refresh loop. Do one successful login, then reload the config entry (or otherwise trigger a second `login_request`) on the same HA session — the second login hits the consent page and fails. With the cookie clear at the top of `login_request`, the reload succeeds instead. That reload/re-login is the cleanest before/after test.
- **A gotcha that fooled me first:** an HA restart gives the shared session a fresh cookie jar, so the *first* login right after a restart succeeds even without the fix — it can look like "works now." The failure only shows on the *second* login within the same run, so testing via reload (not restart) is what makes it reproducible.

`clear_domain` needs aiohttp ≥ 3.8, which HA ships, so that's fine. Happy to open a PR against your branch whenever you'd like it — otherwise I'll leave it with you.

antoinevalentinHA:
Credit where it's due — the shared-session / SSO-cookie diagnosis is your read, not mine; I only flagged that beta fetches the password page differently as a hint. But it lines up: a persisted auth cookie causing the IdP to serve a consent interstitial instead of the password form would drop exactly those markers, and it explains the "first login of the run succeeds, every re-login fails" pattern cleanly.

Clearing the auth-domain cookies at the start of `login_request`, best-effort so a jar quirk can't block login, sounds right. Reproduce it tonight and I'm glad to test the result on an EU account.

Rothgar666:
Went through this tonight and can confirm the whole chain, end to end.

Reproduced it first: one login on a fresh session works fine, then a second login_request on that same session hits the consent page and blows up with the exact "Could not extract CSRF/HMAC/relayState" error. Applied the cookie-clear at the top of login_request (auth-domain cookies wiped before every login attempt, best-effort so a jar quirk can't block login) and re-ran the same two-login sequence — both go through now, second login lands on the password page like the first.

@mstu01 thanks again, the write-up made this trivial to fix, the reload-not-restart gotcha would've cost me a while otherwise.

On the beta question: had a look at request_device_code/poll_device_token there. It's a different shape — the user approves the code manually in their own browser, so it never touches identifier/password pages for the device grant, which is presumably why it doesn't hit this wall. This PR intentionally keeps the login headless (no user step, works for reauth/reload automatically), so I don't think there's a clean way to just merge the two — they're solving the UX tradeoff differently. Happy to talk through converging if that's a priority for master, but for now, but for now I'd rather ship the headless flow with this fix than block on a rewrite.

Should be good for another look.

coreywillwhat:
Thank you so much for your work and contributions! I'm loving seeing how much community is involvement is happening over the last couple months!

The beta resolves a critical issue: US users still need a username/password. I'm not sure how this PR addresses the US username/password requirement. To my knowledge, no QR/Browser/Device code flow currently works in the US (including the VW Group integration), making the beta fix significant.

The beta appears to have reached a stable state; at least for me and seemingly for many others. Given that, it should be promoted to the production/latest release. Until it is, users searching for solutions will keep missing it entirely if they haven't discovered or tested the beta themselves.

Regarding this PR specifically: if it builds on or improves the beta in any way, the right path forward is:

1. Promote the beta to the latest release first.
2. Retarget this PR against that latest release (with the existing beta logic as the baseline).

A lot of work has gone into the beta from multiple contributors, and many fixes have been folded in along the way. Now that the original issue is resolved, the beta should graduate to the latest release. That's what it was working toward. We shouldn't keep merging fixes into beta or master without cutting a new latest release that actually closes the issue the beta was opened to address.

@Kolbi let me know your thoughts. If you're aligned, lets merge the beta into master, and push a latest release.

antoinevalentinHA:
To your point about the US requirement — the beta already handles that, and it's the reason promoting it is safe for US users. Auth is region-gated in `const.py`: `DEVICE_CODE_REGIONS = {REGION_EUROPE}`, so `uses_device_code()` returns true only for the EU. **The US keeps the full username/password flow**; device-code is never forced on a region where it doesn't work. That was the whole point of #791 — device-code for the EU, the untouched v2.2.0 password path for the US, switched by region — and you validated the US side yourself when it went into b3.

So promoting beta → latest doesn't regress the US case; it's the config that resolves it. And this PR (a headless device-code flow) is inherently an EU-side path — no device-code works in the US, as you note — so retargeting it on top of the promoted beta, as an EU enhancement rather than a replacement for the region split, is the clean sequence. +1 on cutting the release.

Kolbi:
@coreywillwhat I am not sure at the moment which direction is most promising.

Beta is working for EU with device-code& USA old way . But this solution here seems to also work for EU without device-code.

Question will be which way will work in the future for EU....

I would vote for: Beta => Master, Rebase this PR and have an option for EU: device-code / non device-code variant as both ways seem to work currently. but I have the feeling one way could be closed in the nearer future....

This PR has not been tested in US?

antoinevalentinHA:
On your "which way survives" instinct, @Kolbi — one structural difference worth weighing: both flows are device-code at the token level, but beta approves the code in a real browser (no page parsing), while this PR automates approval by scraping the email → password → **consent** pages server-side. That HTML surface is exactly what VW keeps changing — the marketing-consent page that broke this PR (and needed the cookie fix) is the tell. So as a hedge, device-code-with-manual-approval is the one least exposed to VW reshaping the login pages; the headless variant is the nicer UX but re-exposes the scraping surface this whole issue was about. Both-as-option is sound — I'd just default to the manual one if forced to pick.

Rothgar666:
@Kolbi to answer directly: no, hadn't tested US, and you were right to ask — turned out this PR had a real gap there. login_request was replacing the login unconditionally, so a US account would've been forced onto the device-code path with no way back to the password flow, and (per @antoinevalentinHA) device-code doesn't exist for US at all. That would've been a straight regression for anyone outside the EU.

Fixed it: login_request is now a small dispatcher, gated by region exactly like beta does it (DEVICE_CODE_REGIONS = {REGION_EUROPE} in const.py, same constant name/shape). EU keeps this PR's headless device-code flow, everything else falls back to the original password/authorization_code flow that was here before I ripped it out. Tested the EU side end to end again (works). Don't have a US account to test against, but ran it through anyway — it correctly pulls the US market config, reaches the password page, and only fails at the point where it's obviously my non-US account being rejected by their side, not a code error. So structurally it should put non-EU users back on the exact path v2.2.0 used.

@antoinevalentinHA — fair point on the scraping surface, and it's hard to argue with the evidence: the exact page this PR broke on is the exact kind of page beta's approach never touches. I don't think that's a reason to drop the headless flow outright, but it's a real argument for not making it the only option.

Which is where I land on @Kolbi's suggestion: beta → master first, then rebase this on top and offer both as a choice for EU (manual approval as default given the fragility argument, headless as an opt-in for people who want it fully unattended) sounds like the right shape to me. I'd rather not bolt a config-flow picker onto this PR blind tonight — that's a bigger change and deserves its own testing pass, and it fits more naturally once beta's device step already exists on master to hang the option off of. Happy to do that rebase once the promotion happens.

jasonbwilliams:
Hello, thanks for the work on this.  It appears this new flow is not working for the CA (Canada) region.  When I use CA it asks for the PIN, then directs me to a username/password prompt, but returns Invalid Credentials.

Rothgar666:
First of all, thank you all — @mstu01, @Kolbi and @antoinevalentinHA — for
taking the time to look at this so carefully. I really appreciate it, and I
clearly should have checked the `beta` branch before opening this.

You're right: `beta` already solves this, and does it better than my PR.
@antoinevalentinHA, thank you for pointing me to #791 — the interactive
device-code grant for EU plus the classic password flow for other regions is
a much cleaner design than what I put together here, and keeping the session
alive with a persisted `refresh_token` avoids the whole problem I ran into.

@mstu01, thank you especially for the detailed diagnosis of the
CSRF/HMAC/relayState failure — that write-up taught me a lot. As I now
understand it, that failure was a consequence of my own approach (re-scraping
the password page on the shared session), and it doesn't apply to `beta`,
which never does that. So the workaround I added isn't something `beta` needs.

Given all that, it doesn't make sense to carry a parallel implementation, so
I'm closing this. If it's useful, I'm very happy to help test `beta` on an EU
account so it can make its way to the stable line for everyone on #775 /
#802 / #805. Thanks again for the patience and the guidance.

Rothgar666:
One thing i forget ...

Closing this in favour of the beta branch — thanks again @mstu01, @Kolbi and @antoinevalentinHA for pointing me there.

I checked out beta and tested the interactive device-code flow end-to-end against my own EU account, and everything worked cleanly:

grafik

- Device authorization + polling — request_device_code returns the user_code / verification URI, I approved it in the browser, and poll_device_token picked up the tokens (access + id + refresh).
- API access — get_vehicle_information returned my vehicle.
- Token refresh — refresh_token_if_necessary refreshes the IDK bearer token and keeps the session alive, with the rotated refresh token persisted via _sync_refresh_token.
- Re-login after restart — login_with_refresh_token restores the session from the stored refresh token (and rotates it), vehicles come back fine afterwards.

So from my side beta is working and looks ready for the stable line — which is what everyone on #775 / #802 / #805 is waiting for.

Two very minor things I noticed while reading through it (non-blocking, just flagging):

1. In AudiConnectAccount.update(), the vehicle-update loop is wrapped in except OSError only. Any other exception during the loop (e.g. a mid-update auth/JSON error) isn't caught there and propagates up. It's still handled one level up by the coordinator (turned into UpdateFailed), so it's harmless in practice — just slightly less precise logging than the OSError path.
2. In add_or_update_vehicle, the first-time add path uses a bare except Exception: pass, which silently swallows errors when a vehicle is first discovered. Not a functional problem, but it can make debugging a failed initial add harder.

Neither is release-blocking. Happy to help test further on EU if useful.

Kolbi:
@coreywillwhat do you want to merge beta into master?
@Rothgar666 feel free to contribute even to the current beta. :) and many thanks for your help/PR - even if it now will not be merged.

coreywillwhat:
@Kolbi I just merged beta into master and cut a new latest release. 

Thanks everyone!

PR #820, opened 2026-08-06
fix: Beta to Master

Kolbi:

Issue #771, opened 2026-07-12
Q6 e-tron API level 1 start_climate_control only works with no request body


WJMarlow:
## Describe the bug

On my Audi Q6 e-tron, `audiconnect.start_climate_control` fails when the integration sends a JSON payload to the API level 1 CARIAD endpoint.

Home Assistant initially reports the action as successful, and the myAudi app briefly shows climatisation as starting. However, after the vehicle synchronises, the request fails and the myAudi app shows an error such as:

- "Could not start air conditioning"
- "Vehicle is not responding"
- Error similar to `E:CV.PA.29`

The myAudi app itself can start climatisation successfully.

I have found a workaround: for my vehicle, the API level 1 climate start request succeeds only when the request body is omitted entirely.

## Vehicle details

- Model: Audi Q6 SUV e-tron
- Year: 2025
- Type: BEV
- Region: EU / UK
- API level: 1
- Integration: Audi Connect HA via HACS
- Lock/unlock works
- Target state of charge works
- Stop climatisation works
- Start climatisation failed until the workaround below
- Start auxiliary heating also failed
- The myAudi app can start climatisation successfully
- The myAudi app no longer appears to offer a temperature-setting option before starting climate control

## Behaviour before workaround

The standard `start_climate_control` action appeared to send climate settings in the request body.

Payloads containing fields such as the following caused the cloud/app to briefly show the request as starting, but then the vehicle rejected it:

    {
      "climatisationMode": "comfort",
      "targetTemperature": 21,
      "targetTemperatureUnit": "celsius",
      "climatisationWithoutExternalPower": true,
      "climatizationAtUnlock": false,
      "windowHeatingEnabled": false,
      "zoneFrontLeftEnabled": false,
      "zoneFrontRightEnabled": false,
      "zoneRearLeftEnabled": false,
      "zoneRearRightEnabled": false
    }

I tested several variants:

1. Original action with temperature and other options
2. Removing `targetTemperature` and `targetTemperatureUnit`
3. Sending only:

    {
      "climatisationMode": "comfort"
    }

4. Adding app-style headers such as:
   - `Accept: application/json`
   - `User-Agent`
   - `X-App-Version`
   - `X-App-Name: myAudi`

All of those still failed.

## Working workaround

Changing the API level 1 branch of `start_climate_control` to send no request body works.

The working logic is:

    elif api_level == 1:
        headers = {
            "Authorization": "Bearer "
            + self._bearer_token_json["access_token"],
[...]

PR #822, opened 2026-08-06
fix: start climatisation with the vehicle's own settings when none are given (#771)

its-me-prash:
### What this fixes

On API level 1 the integration always sends a full climatisation settings payload. Vehicles without a comfort climatisation profile reject it. The rejection only arrives after the car has synchronised, so Home Assistant reports the action as successful and the error surfaces in the myAudi app instead, which is what makes this one awkward to spot.

Reported in #771 for a Q6 e-tron, with a second reporter in the same thread independently arriving at the same workaround and confirming it works.

### What changed

When the service call supplies no settings at all, send no body and let the vehicle apply the settings it already holds. That is what the myAudi app does on a car with no comfort climatisation profile. Any supplied setting still produces the full payload exactly as before.

Two smaller things came out of the same code path:

- `climatisationMode` was serialised as a literal `null` whenever the service call left it unset, since the value is passed straight through. It now falls back to `comfort`.
- `audi_account.py` no longer substitutes `False` for the unset booleans. Without that, the service layer cannot tell "nothing was asked for" from "everything was explicitly asked to be off". API level 0 is unaffected: its payload is built with an explicit `bool()` conversion and still sends every field.

### Behaviour change

`start_climate_control` called with no parameters previously imposed 21 degrees with all zones off. It now leaves the vehicle's own settings alone. Callers that pass settings see no change.

### Tests

`tests/test_climate_start_payload.py`, six cases: the empty body, a supplied temperature, a single supplied setting, the mode fallback, an explicit all off, and an API level 0 regression guard. Full suite is 10 green and needs no network.

### Verification

I do not have a Q6 e-tron, so this rests on the two independent reports in #771 rather than my own live test. Worth a prerelease check with an affected car before it reaches stable.

### Target branch

Opened against master because beta no longer exists after #820. Happy to retarget as soon as it is back.

### DIFF
--- a/custom_components/audiconnect/audi_account.py
+++ b/custom_components/audiconnect/audi_account.py
@@ -211,16 +211,19 @@ class AudiAccount(AudiConnectObserver):
 
     async def start_climate_control(self, vin: str, service: ServiceCall) -> None:
         """Start climate control for a vehicle by VIN."""
+        # Pass the fields through as supplied, without substituting defaults, so
+        # the service layer can tell "the caller wants everything left alone"
+        # apart from "the caller explicitly asked for all of it off".
         await self.connection.start_climate_control(
             vin.lower(),
             service.data.get(CONF_CLIMATE_TEMP_F),
             service.data.get(CONF_CLIMATE_TEMP_C),
-            service.data.get(CONF_CLIMATE_GLASS, False),
-            service.data.get(CONF_CLIMATE_SEAT_FL, False),
-            service.data.get(CONF_CLIMATE_SEAT_FR, False),
-            service.data.get(CONF_CLIMATE_SEAT_RL, False),
-            service.data.get(CONF_CLIMATE_SEAT_RR, False),
-            service.data.get(CONF_CLIMATE_AT_UNLOCK, False),
+            service.data.get(CONF_CLIMATE_GLASS),
+            service.data.get(CONF_CLIMATE_SEAT_FL),
+            service.data.get(CONF_CLIMATE_SEAT_FR),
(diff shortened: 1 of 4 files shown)


WJMarlow:
Tested this successfully on my Audi Q6 e-tron affected by #771.

I installed the current beta (2.3.2b1) and confirmed that the audi_services.py in use was the unmodified beta version containing this PR (Git blob SHA 91670a221e0119f2a172f22db76489a4d90c286b). I then restarted Home Assistant Core. This matches the merged #822 implementation.

I called audiconnect.start_climate_control with only the device ID and no temperature, mode, seat, glass-heating or other climate settings.

Debug logging confirmed that the new code path was being used:

Attempting to start climate control with API Level 1 and country DE.
No climate settings supplied; starting climatisation with the settings stored in the vehicle.

Result: the fix works on the physical Q6.

Climatisation starts successfully in the car.
The myAudi app shows climatisation running.
The HVAC is physically operating in the vehicle.
It continues running normally rather than stopping after a few seconds.
On my latest test it was still running normally after 15 minutes.

On one earlier test I did get a delayed myAudi “vehicle communication” error (E:CV.PA.29), despite the climate having started successfully and continuing to run afterwards. The popup appears to be intermittent: on the subsequent 15minute test there was no error at all.

So from my testing, #822 fixes the actual Q6 climate-start problem. The occasional delayed E:CV.PA.29 appears to be a separate/intermittent acknowledgement or status-sync issue rather than failure of the start command.

Thanks for implementing this — happy to test any further changes on the Q6 if useful.

PR #833, opened 2026-08-18
feat: Beta to Master

Kolbi:
### Bug Fixes

* **Improved climate control behavior:** When starting climatisation without specifying temperature, seat heating, window heating, or other climate settings, the integration now uses the settings stored in the vehicle instead of sending default values. This improves compatibility with vehicles such as the Q6 e-tron. (#822)
* **Improved login retry handling:** Failed login attempts now use exponential backoff instead of retrying at a fixed interval, reducing unnecessary requests to the Audi service during outages or temporary authentication issues. (#823)
* **Improved handling of unknown vehicle actions:** Unknown actions are now logged instead of being silently ignored. (#824)
* **Improved lock/unlock error handling:** Failed vehicle lock or unlock commands now raise a Home Assistant error instead of appearing to succeed silently. (#825)
* **Home Assistant 2026.12 compatibility:** Reauthentication and reconfiguration now explicitly schedule the integration reload, replacing the deprecated reload behavior. (#828)

### Maintenance

* Updated pre-commit dependencies and development tooling. (#827, #832)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.16.2
+    rev: v0.16.3
     hooks:
       - id: ruff-check
         args:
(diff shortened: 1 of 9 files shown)


Kolbi:
I merged it since around 25 users are already using v.2.3.2b*.

Issue #790, opened 2026-07-21
Start Charger Doesn't Wait for Vehicle Refresh


coreywillwhat:
**Describe the bug**
When requesting the charger to start timed charger or immediate charging, the success message comes  immediately and the cloud refresh delay of 5 seconds is not enough to let the vehicle report the update. The result is that the data for the `charging mode` sensor is inaccurate and doens't display the change that was just requested.

```yaml
2026-07-21 10:59:02.399 DEBUG (MainThread) [custom_components.audiconnect.audi_connect_account] Sending command to start charger to vehicle XXXXXXXXXXXXXX
2026-07-21 10:59:03.261 DEBUG (MainThread) [custom_components.audiconnect.audi_api] [VCF-RATE-LIMIT] Remaining calls: 45
2026-07-21 10:59:03.262 DEBUG (MainThread) [custom_components.audiconnect.audi_api] Returning JSON data: {'data': {'requestID': '03b5b470-7d38-44f6-aa27-16f0adec04f9'}}
2026-07-21 10:59:03.262 DEBUG (MainThread) [custom_components.audiconnect.audi_connect_account] Successfully started charger of vehicle XXXXXXXXXXXXXX
2026-07-21 10:59:03.262 DEBUG (MainThread) [custom_components.audiconnect.audi_account] Sleeping 5 seconds before cloud data refresh
2026-07-21 10:59:08.264 DEBUG (MainThread) [custom_components.audiconnect.audi_account] Starting refresh cloud data...
```

**Your Vehicle Details**
Model: etron
Year: 2021
Type (Gas/Hybrid/Electric): Electric
Region (EU/US/CA/CN): US

antoinevalentinHA:
I had a look at this — I think the delay is a symptom rather than the cause: `set_battery_charger` never waits for anything.

The `PUT` to `charging/mode` returns a `requestID` (visible in your log: `{'data': {'requestID': '03b5b470-…'}}`), but the response isn't captured, so the method returns straight away, the account layer logs "Successfully started charger", and the 5s sleep + refresh fires while the car still hasn't applied the change. Lengthening `update_sleep` would paper over it — it'd work on a fast car, still miss on a slow one, and waste time when the car is quick.

[...]

Issue #765, opened 2026-06-25
Target state of charge not updated


RWSF74:
**Describe the bug**
Target state of charge is not updated anymore (sensor.audi_e_tron_55_quattro_target_state_of_charge in my case)

**To Reproduce**
Reloading integration does not update either

**Expected behavior**
With each polling this sensor should be updated, if changed. Other sensors are updated correctly.

**Screenshots**

**Logfile**
How to enable audiconnect debugging?
Settings > Integrations > Audi Connect > Enable Debug Logging
Run service refresh_cloud_data
Disable Debug Logging

**Your Vehicle Details**
Model: Audi etron 55
Year: 2020
Type (Electric):
Region (EU):

woopstar:
I am seeing the same behavior.

woopstar:
@coreywillwhat

sla004:
I have also same problem

PR #834, opened 2026-08-18
feat: expose the charging profile that governs, and make start/stop actually charge

### LINKED ISSUE #722: Position specific target state of charge
Audi e-tron 2022

In car it's possible to set location specific target state of charge. When car is parked in this location, HA detects that correctly, but target state of charge is not changed. It still shows "general" or default target state.

### LINKED ISSUE #725: Action "Stop Charger" deprecated
The "Stop Charger"  Vehicle action does not work .

The Debug error output of HA audiconnect shows the reason 

2026-04-15 19:45:35.489 DEBUG (MainThread) [custom_components.audiconnect.audi_connect_account] Sending command to stop charger to vehicle wauxxxxxxxxxxx
2026-04-15 19:45:35.489 ERROR (MainThread) [custom_components.audiconnect.util] Unable to stop charger of vehicle wauxxxxxxxx: The 'Stop Charger' service is deprecated and will be removed in a future release.

Model: Audi A3 TFSI e
Year: 2026
Type : Hybrid
Region EU

lmjabreu:
## Overview

The location charging profile decides where a charge stops, and the global `charging/settings` target does not. The two disagree whenever a profile is in force, and the integration only ever showed the global one.

Measured on a 2022 Q8 Sportback e-tron parked at a profile named "Home":

| | |
|---|---|
| `chargingProfiles` → profile 1 "Home" `targetSOC_pct` | **60** |
| `charging.chargingSettings.value.targetSOC_pct` | 70 |
| `sensor.*_target_state_of_charge` | 70 |
| where the car actually stopped | **60** |

Setting the global target to 65 while the car sat at 68% did nothing. Setting the Home profile to 60 stopped the charge immediately. That is #722, and it is a likely explanation for reports like #765 where the target sensor looks stale or wrong: it may simply not be the number in force.

`chargingProfiles` and `chargingTimers` are already fetched by `get_stored_vehicle_data` and thrown away without being parsed.

Separately, none of the three charger actions started or stopped a charge. `start_charger` and `start_timed_charger` `PUT charging/mode`, which is a mode preference, and `stop_charger` raised `NotImplementedError` (#725). Because `AudiConnectAccount.set_battery_charger` logs its exceptions and returns `None` instead of propagating, that raise never reached the caller, so the service answered 200 while nothing reached the car.

## Changes

**Charging profiles and timers become entities.** Six sensors: the active profile's target SoC and name, its min SoC, the preferred charging window start and end, and the count of enabled timers. The target sensor carries `profile_id`, `profile_name`, `global_target_soc_pct` and the full profile list as attributes, so the disagreement is visible in one place.

Profile `position` is deliberately left out of attributes. It is the owner's home coordinates, and attributes land in the recorder database and in diagnostics downloads.

**`start_charger` and `stop_charger` now start and stop a charge**, via `POST charging/start` and `POST charging/stop`, keeping the combined `charging/start-stop` route as a 404 fallback for older vehicles. Both routes are used by `tillsteinbach/CarConnectivity-connector-volkswagen` and `its-me-prash/vwgroup-connect-ha` against the same BFF.

**Commands are confirmed against `pendingrequests`**, bounded to 3 polls 5s apart rather than the 10 x 10s `check_bff_request_succeeded` uses, because each poll spends a call from the metered daily allowance and blocks the service call. Running out of attempts logs a warning and still reports success, deliberately: a command that did work must never be reported as broken. An explicit rejection raises. This partially addresses #790.

**New `set_charge_mode` service** (`manual` / `timer`).

**The three charger actions raise `HomeAssistantError` when the command did not take**, so a caller can tell success from silence.

## Behaviour change worth flagging

`start_charger` no longer sets the mode to `manual` as a side effect, because it now performs the action instead of expressing a preference. Anyone who relied on that side effect should call the new `set_charge_mode` service. `start_timed_charger` is unchanged and still selects timer mode.

Happy to mark this `feat!` if you would rather signal that in the release notes.

## Reference

- Fixes #725 (Stop Charger deprecated)
- Fixes #722 (position specific target state of charge)
- Partially addresses #790 (start charger does not wait) and #765 (target state of charge not updated)

## Test plan

Automated, `pytest tests/`: 14 new tests over a real `selectivestatus` payload captured from the car, with the home coordinates scrubbed and every other value verbatim. Run against unmodified `master`, 12 of the 14 fail, so the suite is known to be able to fail.

To reproduce the reported problem and verify the fix on a car with a location profile:

1. In the myAudi app, set the charging profile for your current location to a target **below** the vehicle's current SoC, and set a different global target.
2. Reload the integration and compare `sensor.*_target_state_of_charge` with `sensor.*_active_charging_profile_target_soc`.
   - [ ] Before this change: only the global target is available, and it does not match where the car stops.
   - [ ] After: `active_charging_profile_target_soc` shows the profile value, and its `global_target_soc_pct` attribute shows the other one.
3. Plug in and call `execute_vehicle_action` with `stop_charger`.
   - [ ] Before: the log shows `The 'Stop Charger' service is deprecated`, and the service still returns success.
   - [ ] After: the charge stops. The debug log shows a `requestID`, and `pendingrequests` reports `path /vehicles/{vin}/charging/stop`, `operation stop`.
4. Call `start_charger` while plugged in and below the profile target.
   - [ ] The charge starts.

Verified against a live car for steps 2 and 3, and for the acceptance half of step 4: both routes returned a request id and `pendingrequests` echoed the expected path and operation. The physical start in step 4 was not observed, because that car was above its governing profile target at the time and so declined to start. Worth a second pair of eyes on a car that is below target.

Vehicle state stays stale for minutes after a command, so judge the outcome from actual charging power rather than `charging_state` read immediately afterwards.


lmjabreu:
@its-me-prash would you have time to look at this one? I took `charging/start` and `charging/stop` from `vwgroup-connect-ha`, so you're better placed than most to say whether I've got it right for Audi.

Both are verified on a Q8 e-tron: accepted with a request id, and `pendingrequests` echoed back `path /vehicles/{vin}/charging/stop`, `operation stop`. The part I'm unsure about is the `charging/start-stop` fallback I kept behind a 404. Have you seen a car that actually needs it, or is it dead weight?

The other judgement call is that `start_charger` no longer sets the mode to `manual` as a side effect, now that it performs the action instead. `set_charge_mode` covers the mode. Tell me if that split is wrong.

One for whoever picks this up: the Validate workflow is sitting at `action_required`, so hassfest and HACS validation haven't run on the branch yet.

its-me-prash:
Hey @lmjabreu, took a look, and yeah, you got it right for Audi. Audi EU and VW EU sit on the same CARIAD BFF, so the charging routes are identical. We run both brands off the same code path and start/stop works on both, so nothing Audi specific to worry about there.

On the charging/start-stop fallback, I actually just pulled the current myAudi app and had a look. It only uses the separate charging/start and charging/stop, the combined start-stop route is not in the app at all. So for current Audi firmware it really is dead weight, you are not missing anything by keeping it behind a 404. We carry it ourselves purely as insurance for older model years, and honestly I do not have a single confirmed car that ever needed it. If you want to drop it, low risk. I would personally keep it one more release just in case a very old MY 404s the separate route, but the app itself has fully moved to the split routes, so your instinct is right.

If you do keep it, one heads up from our own implementation: the combined start-stop route wants the action in the body, so {"action": "start"} or {"action": "stop"}, not an empty body. The separate routes take an empty body, the combined one does not. If the fallback posts empty it'll 200 and do nothing, which is the worst kind of silent no-op. We learned that one the hard way.

The mode split is right. Start should perform the action, not express a preference, and the mode belongs in its own service. That's exactly how we have it, and pulling the manual side effect off start is the correct call.

On the profiles, good catch, and we landed in the same place. The location profile is what actually governs where the charge stops, not the global charging settings target, and the two disagree constantly. We already ship the active profile target and its name as their own sensors sitting right next to the global one, for exactly the reason you measured (profile says 60, global says 70, car stops at 60). So your read on that is solid, and leaving the profile position out of the attributes is the right instinct too.

Nice PR overall, and good that you verified it on a real Q8 rather than just trusting the routes, the request id plus the pendingrequests echo is exactly the proof I'd want to see. The 12 of 14 failing on master is a nice touch too, shows the tests actually bite. The pendingrequests confirmation and raising on the silent failures are both the right calls.

Unrelated to the review, one ask. That scrubbed Q8 e-tron selectivestatus payload you captured is exactly the kind of real Audi PPE sample our test suite is thin on, most of our real fixtures are Skoda and SEAT/CUPRA. Would you mind if I pulled it into vwgroup-connect-ha as a test fixture, credited to you? It'd let me pin the same charging profile parsing against a real car on our side too, and it's a fair trade since I borrowed the start/stop routes off you here. Completely fine if you'd rather not, it's your car's capture.

lmjabreu:
Thanks for the feedback.

Empty body: I checked the fallback and it does send `{"action": "start"}` / `{"action": "stop"}`. Thanks for flagging as we're getting rid of a NOOP 200.

I think I'll keep the combined route for now for the reason you gave (insurance in case a very old model year 404s the separate routes).

Feel free to take the fixture, no need to credit even, happy to contribute more if helpful.

FYI: I scrubbed a real location from one of the profile names and force-pushed, renamed it to Work, the coordinates were already scrubbed. The rest is unchanged.

The fixture is from a 2024 Q8 Sportback e-tron 55, EU. Worth knowing if you pin tests against it: `departureProfiles` and `departureTimers` don't come back at all on this car, along with `auxiliaryHeating`, `batterySupport`, `lvBattery`, `oilLevel` and `readiness`. Also `chargingProfiles` carries its own `carCapturedTimestamp`, separate from the `charging` block, about 8 minutes apart in this capture.

lmjabreu:
One more thing that might be useful given you ship the profile target sensors. There is a write for it, which I don't think is in any of the open source clients. `PUT {bff}/vehicle/v1/vehicles/{vin}/charging/profiles`, body `{"profile": {<the whole profile object, one field changed>}}`. It replaces the entire profile so it has to be read-modify-write, and the id in the body picks which profile, independent of `vehiclePositionedInProfileID`. Also a POST to create and a DELETE on `charging/profiles/{id}`, and the timers equivalent is `PUT /vehicle/v2/departure/{vin}/charging/timers` with `{"timers": [...]}`.

I got it out of the myAudi app and tested it against the Q8. It's in a follow-up PR I'll open here once this one lands, since it depends on the confirmation helper in this PR.

lmjabreu:
Forgot to link it: https://github.com/lmjabreu/audi_connect_ha/pull/2. It's a draft in my fork based on this branch, so the diff is just the profile write on top. Feel free to lift anything from it.

lmjabreu:
Follow-up is now open: #839, rebased onto `dev`. Closed the fork one, that link is stale.

PR #837, opened 2026-08-22
feat: add a Position Last Updated sensor

lmjabreu:
## Overview

`device_tracker` reports where the car is, but not how old that reading is. `parkingposition` carries its own `carCapturedTimestamp` and only moves when the car reports a new position, so the tracker can show a confident `home` that is hours stale, and nothing distinguishes it from a live fix. An automation that gates on vehicle location has no way to tell the difference.

`update_vehicle_position` already parses that timestamp and stores it in `state["position"]["timestamp"]`. It was simply never surfaced. This exposes it as a diagnostic timestamp sensor so callers can decide what counts as too old, rather than trusting the position blindly.

Concrete case this came from: a panel gating a persistent setting on "is the car home" acted on a position that turned out to be many hours old. The coordinates were right when captured and wrong to act on later, and nothing in Home Assistant made that visible.

## Changes

Adds `position_last_updated` (plus its `_supported` partner) to `AudiConnectVehicle`, and a `Position Last Updated` sensor with `SensorDeviceClass.TIMESTAMP`, in the diagnostic category.

No new API call and no extra request against the rate limit: the value is already in the response the integration fetches. The entity is only offered when the vehicle actually reports a timestamp, since `update_vehicle_position` stores `None` when `carCapturedTimestamp` is absent.

## Reference

n/a. No linked issue; found while building automations on top of the position data.

## Changelog

Added a Position Last Updated sensor reporting when the vehicle last reported its position.

## Test plan

Automated, `pytest tests/`: 8 tests over the real `AudiConnectVehicle` property, covering a normal capture, an absent timestamp, no position at all, a missing key, a week-old capture, and timezone-awareness (a naive datetime would break `SensorDeviceClass.TIMESTAMP` at runtime). They fail against `master`, so the suite is known to exercise the change.

To verify on a vehicle:

1. Note the current value of `sensor.<vehicle>_position_last_updated`.
   - [ ] It matches when the car last parked, not when Home Assistant last polled.
2. Call `audiconnect.refresh_cloud_data` without moving the car.
   - [ ] The sensor does not change, because the vehicle has not reported a new position.
3. Drive the car and park somewhere new, then refresh.
   - [ ] The sensor advances to the new capture time.
4. On a vehicle that does not report `carCapturedTimestamp`.
   - [ ] The entity is not created, and the other position entities are unaffected.

### DIFF
--- a/custom_components/audiconnect/audi_connect_account.py
+++ b/custom_components/audiconnect/audi_connect_account.py
@@ -2062,6 +2062,22 @@ class AudiConnectVehicle:
         return self._vehicle.state.get("isMirrorHeatingActive") is not None
 
     @property
+    def position_last_updated(self):
+        """When the vehicle last reported the position, not when HA last polled.
+
+        parkingposition carries its own carCapturedTimestamp and only moves when
+        the car reports a new position, so a location can be hours old while the
+        entity looks current. Exposing the capture time lets an automation decide
+        whether a position is fresh enough to act on.
+        """
+        if self.position_last_updated_supported:
+            return (self._vehicle.state.get("position") or {}).get("timestamp")
+
+    @property
+    def position_last_updated_supported(self):
+        return (self._vehicle.state.get("position") or {}).get("timestamp") is not None
+
+    @property
     def park_time(self):
         if self.park_time_supported:
             return self._vehicle.state.get("vehicleParkingClock")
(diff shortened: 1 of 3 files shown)


Kolbi:
@lmjabreu just checked it now in system do you have the icon: "mdi:map-marker-clock" in your system per default?
https://github.com/audiconnect/audi_connect_ha/blob/8b49137a163aae67e050cbb56307d2fec1ca0d35/custom_components/audiconnect/sensor.py#L404

Issue #840, opened 2026-08-27
Refresh rejected with invalid_client triggers a reauth prompt that reauth cannot fix


antoinevalentinHA:
### Summary
When the IDK refresh-token exchange is rejected, the integration raises `ConfigEntryAuthFailed` for **any** rejection, which prompts the user to re-authenticate. But an `invalid_client` rejection cannot be fixed by re-authenticating — the `client_id` is hardcoded, so a reauth reuses the exact same client — and these rejections are often transient. The result is a scary, un-actionable reauth prompt for a problem that clears itself.

### Environment
- Version: v2.3.1 at capture time; behavior unchanged on **v2.4.0** (`login_with_refresh_token` raises a generic `AudiAuthError`; `async_refresh_data` escalates every `AudiAuthError` to `ConfigEntryAuthFailed`).
- Region: EU (DE), device-code flow.

### What happened (2026-08-26, a VW token-instability window)
```
16:29  coordinator: Authentication failed while fetching audiconnect data:
       IDK refresh rejected: invalid_client        -> reauth banner appears
19:33  audi_connect_account: Unable to start climate control ... 401
       .../fs-car/bs/climatisation/v1/Audi/DE/vehicles/WAUZZZ.../climater/actions
19:35  VEHICLE REFRESH: ... Request ... timed out
```
All three were transient: by the next morning entities were refreshing again, and a plain integration **reload** cleared the stale reauth banner without ever touching VW's (currently flaky, #835) verification page. So the reauth prompt was never actionable and never necessary.

### Root cause
`login_with_refresh_token` raises the same `AudiAuthError` whether the OAuth error is `invalid_grant` (the refresh token itself is dead -> reauth **is** the fix) or `invalid_client` / a transient/malformed response (reauth **cannot** help). `async_refresh_data` then escalates all of them to `ConfigEntryAuthFailed`.

### Proposed fix
Distinguish the OAuth error on a refresh rejection:
- **`invalid_grant`** -> `ConfigEntryAuthFailed` (reauth genuinely needed). Unchanged.
- **`invalid_client` / transient / malformed** -> a normal failure the coordinator treats as `UpdateFailed` (retryable). The next poll usually recovers, with no reauth prompt.

This would have made the 2026-08-26 window a non-event.

### Twin symptom (same family, separate fix)
[...]

PR #843, opened 2026-09-01
fix: only prompt reauth when a refresh is rejected with invalid_grant (#840)

antoinevalentinHA:
Addresses #840.

A rejected IDK refresh raised the same `AudiAuthError` regardless of the OAuth error, and `async_refresh_data` escalates every `AudiAuthError` to `ConfigEntryAuthFailed`. So a transient VW blip or an `invalid_client` — which a reauth **cannot** fix, since the `client_id` is fixed in code — produced an un-actionable reauth prompt for a problem that clears itself.

### Fix

Following the inverse-rule shape @its-me-prash suggested on the issue — list the **hard** failures, not the transient ones — scoped to the refresh exchange:

- **`invalid_grant`** (the refresh token itself is dead) → `AudiAuthError` → reauth. Unchanged.
- **everything else** (`invalid_client`, a transient body, a malformed response) → a new retryable `AudiTokenRefreshError`, which the coordinator surfaces as `UpdateFailed` and retries next cycle (bounded by the existing login backoff). An unknown code degrades to a retry — the harmless direction.

Applied at both refresh points (`login_with_refresh_token` and the periodic IDK refresh).

| refresh error | before | after |
| --- | --- | --- |
| `invalid_grant` | reauth prompt | reauth prompt |
| **`invalid_client`** | **reauth prompt (can't help)** | **retry** |
| transient / malformed | reauth prompt | retry |

### Scope

`invalid_grant` is the only hard failure that surfaces at a *refresh* exchange; the other user-actionable reasons (T&C, consent, 2FA, unfinished portal, bad credentials) surface at the initial login / device approval. Extending the same inverse rule across the whole login path is a sensible follow-up. The twin symptom in #840 — a `401` on an MBB action from an expired `vwToken` with no re-mint/retry — is also left for a follow-up.

### Testing

Unit-tested (`tests/test_refresh_reject_classification.py`): `invalid_grant` → `AudiAuthError`; `invalid_client` / `server_error` / malformed → `AudiTokenRefreshError`; and the retryable error is not an `AudiAuthError` subclass (or `async_refresh_data` would still escalate it). `audi_services` has no Home Assistant import, so this runs in the existing harness-free test setup.

*Prepared with AI assistance; reviewed by me.*

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -10,7 +10,7 @@ import re
 import uuid
 from datetime import datetime, timedelta, UTC
 from hashlib import sha256, sha512
-from typing import Any
+from typing import Any, NoReturn
 from urllib.parse import parse_qs, urlencode, urlparse
 
 from aiohttp import ClientResponseError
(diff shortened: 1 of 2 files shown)


its-me-prash (commented):
Looks right to me, and the narrower scope is the better call. `invalid_grant` really is the only one that still makes sense as a hard failure at a refresh exchange, and keeping the escalation for it means the one actionable case still gets through.

The test pinning `AudiTokenRefreshError` as not being an `AudiAuthError` subclass is the one I would have asked for. That inheritance is exactly what would quietly undo the whole thing a year from now.

One gap, and it is a third refresh site this PR does not touch. In `_finalize_session`, immediately below the mbboauth exchange you hardened in #792, the mbboauth refresh does:

```python
self.vwToken = json.loads(mbboauth_refresh_rsptxt)
```

with no check at all. If that refresh is rejected, the error body is stored as the token, which is precisely the failure your own comment a few lines above warns about, surfacing much later as `KeyError('access_token')` in `get_tripdata()` or a `TypeError` in `get_climater()`. So the auth step is guarded and the refresh step directly underneath it is not, and it is the same classification question you are solving here.

Whether that belongs in this PR or alongside the MBB `401` work is your call. Pulling it in would make the rule cover every refresh path; leaving it out keeps this one tight and easy to review. Either way the change as it stands is a clear improvement.

antoinevalentinHA:
Good catch — pulled it in. There were actually **two** unguarded `self.vwToken = json.loads(...)` sites (the `_finalize_session` one you pointed at, plus the periodic vwToken refresh), both storing the error body as the token exactly as the #792 guard one line up warns about. Both now route through the same `_raise_refresh_rejected`, so the rule covers every refresh path: `invalid_grant` → reauth, everything else → retryable.

mbboauth returns `error_description` rather than a bare `error`, so I made the classifier prefer it for the message (matching your #792 guard) while still classifying on `error` — an mbboauth body with no `error` field degrades to retryable, the safe direction. Test extended for that shape.

Thanks for the review.

Kolbi:
@its-me-prash @coreywillwhat @antoinevalentinHA I am currently in vacation & will have a look into all PRs after it.

Issue #829, opened 2026-08-11
Cannot connect to host ha-5a.prd.nar.vwg.vwautocloud.net: CERTIFICATE_VERIFY_FAILED


cavedon:
**Describe the bug**
Every about 5-6 hrs, all the data becomes "Unavailable" and the Audi integration wants me to re-authenticate. "Your Audi sign-in is no longer working. Enter your credentials again."
After I re-authenticate, it works for another 5-6 hrs.

It used to work fine until some time last week.

I see an error:
```
2026-08-11 10:28:25.618 ERROR (MainThread) [custom_components.audiconnect.audi_connect_account] TRIP DATA: An unexpected error occurred while updating trip data for VIN: *************1120. Error: Cannot connect to host ha-5a.prd.nar.vwg.vwautocloud.net:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1082)')]
```

Interestingly, after running the refresh_cloud_data action, the data was fetched correctly and is no longer unavailable, but the request to authenticate remains.

**To Reproduce**
1. Enter my credentials in the Audi integration
2. Wait 5-6 hrs
3. All data becomes Unavailable, and the Audi connect integration wants to reauthenticate

**Expected behavior**
No need to reauthenticate.

**Screenshots**

<img width="1080" height="685" alt="Image" src="https://github.com/user-attachments/assets/1cae67a4-0a3d-499c-8a8c-1ecbb249b918" />

**Logfile**
[home-assistant_audiconnect_2026-08-11T17-28-35.671Z.log](https://github.com/user-attachments/files/30945613/home-assistant_audiconnect_2026-08-11T17-28-35.671Z.log)

**Your Vehicle Details**
Model: A6 Sportback e-tron
Year: 2026
Type (Gas/Hybrid/Electric): Electric
Region (EU/US/CA/CN): US

cavedon:
Seems to be the same as #714, which was automatically closed last May

Kolbi:
@coreywillwhat it's related to US any idea?

its-me-prash:
@Kolbi to your question about this being US related: it is, but only halfway. EU accounts hit the same thing, and the reason for both is in `_fill_home_region`.

[...]

PR #844, opened 2026-09-01
fix: never route to the vwautocloud home region host (#829)

its-me-prash:
### What this fixes

The home-region lookup can return a `ha-5a.prd.<region>.vwg.vwautocloud.net` base URI. That host does not present a usable certificate chain, so anything routed there fails with `CERTIFICATE_VERIFY_FAILED`.

`_fill_home_region` already knew about this. Its comment says the endpoint returns that host and that it is not a valid endpoint, which is why the region is set statically instead. But the guard reads:

```python
if self._country.upper() != "US" and self._api_level == 1:
```

so it only applies to non-US accounts on API level 1. Every other combination falls through to the lookup and adopts whatever comes back, including the host the comment just declared invalid.

That accounts for both reports:

- #829, a US account, fails the `!= "US"` half and gets `ha-5a.prd.nar...`
- the report in #772 (region DE, API level 0) fails the `api_level == 1` half and gets `ha-5a.prd.eu...`

Two regions, one gap.

### What changed

Gate on the returned value rather than on the account. If the base URI points at `vwautocloud`, discard it and keep the static default that already works, whatever the country or API level. A usable base URI is still adopted exactly as before, and the existing shortcut for non-US API level 1 is untouched.

### Tests

`tests/test_home_region_fallback.py`, five cases: the US path, the API level 0 Europe path, a usable base URI still being adopted, a failed lookup leaving the default in place, and the existing non-US level 1 shortcut still returning early. Without the change the first two fail and the other three pass, so they pin the actual behaviour rather than the implementation. Full suite is 58 green and needs no network.

### Verification

I have no account in either affected configuration, so this rests on the two reports plus the endpoint behaviour the existing comment already documents. What the change cannot do is fix a certificate problem on the user's own side: `self-signed certificate in certificate chain` can also come from TLS interception by a router or antivirus. It removes the case where we route to a host the code already considers unusable, which is what both reporters hit.

### DIFF
--- a/custom_components/audiconnect/audi_services.py
+++ b/custom_components/audiconnect/audi_services.py
@@ -24,6 +24,12 @@ from .util import get_attr, to_byte_array
 MAX_RESPONSE_ATTEMPTS = 10
 REQUEST_STATUS_SLEEP = 10
 
+# The home-region lookup can return a ha-5a.prd.<region>.vwg.vwautocloud.net base
+# URI. That host does not present a usable certificate chain, so every call routed
+# there dies with CERTIFICATE_VERIFY_FAILED. Anything containing this is discarded
+# in favour of the static default.
+UNUSABLE_HOME_REGION_HOST = "vwautocloud"
+
 # Charging start/stop confirmation. Kept short because each poll spends one
 # call from the metered daily allowance and blocks the service call.
 CHARGING_CONFIRM_ATTEMPTS = 3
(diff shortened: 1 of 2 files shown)


antoinevalentinHA:
Had a look — correct, and the test coverage is the right shape. One thing I double-checked that holds up: the account guard at the top of `_fill_home_region` isn't made redundant by the value-based check, because it sets a *different* default (`mal-3a…`) for the non-US / API-level-1 case rather than the general `mal-1a` / `msg` one. Dropping it in favour of the value check alone would quietly move those accounts onto the wrong default — good that you kept it, and `test_non_us_api_level_1_keeps_its_existing_shortcut` pins exactly that. Nice fix.

PR #849, opened 2026-09-05
feat: control climatisation with a climate entity

lmjabreu:
> Rebased onto current `dev`, and independent of everything else open. The diff is this change alone.
>
> It also carries the fix for a bug it causes, which used to sit several PRs downstream in #856. See the thread below.
>
> **Merge order.** Three independent lines, so nothing here needs coordinating across them:
>
> - #855, on its own
> - #848 → #852
> - #849 → #853 → #856, then #857 (draft) last
>
> Every one of these edits `readme.md`, so each merge makes the others conflict. That is expected rather than a problem with the PR. I rebase and push as soon as I see it, normally the same working day, so a conflicting state here is stale rather than stuck.

## Overview

Climatisation needs a target temperature to start, which is what a climate entity is for. This is why it was left out of the controls PR rather than shipped as a switch that could not carry one.

State comes from the car. `climatisation_state` gives the mode, so heating, cooling and ventilation map to the matching `HVACAction`, and a running mode the map does not name reads as running rather than idle. The target temperature is held locally here and restored across restarts.

**Correction, after this was first written:** the car does report the target temperature, as `targetTemperature_C` in `climatisation.climatisationSettings`. That block is fetched on every poll and dropped, so the local value in this PR is a stand-in for a reading that was available all along. A follow-up parses the block and makes the entity prefer the car's value, keeping the local one only as a fallback for a vehicle that reports none. Worth naming plainly, because it is the same mistake the table below flags in #751: asserting a field is absent without checking the payload.

Two omissions are deliberate. There is no current temperature: the car reports the outdoor reading, not the cabin, and presenting that as a climate entity's current temperature would misstate what it measures. Seats, glass heating, climatisation-at-unlock and the comfort/economy mode are not exposed either, because the vehicle never reports them back. Those stay on `start_climate_control`.

#751 also adds a climate entity and has been open since June. I read it while writing this and it is where the approach came from, but it reads three attributes that do not exist, so each of those has a test here:

| #751 | Here |
| --- | --- |
| reads `climatisation_active`, which no vehicle defines, so its mode only reflects the last command | reads `climatisation_state` |
| reads a `temperature` property, which no vehicle defines, as current temperature | no current temperature |
| resets the target to 21.0 on every restart | restores it (and see the correction above: the car reports it, and the follow-up reads it) |
| creates the entity for every vehicle | gates on climatisation support |
| sends `temp_f` and `temp_c`, and the service prefers `temp_f`, so the 0.5 step is quantised through a Fahrenheit round trip | sends only `temp_c` |

## Reference

Related: #751. Standalone, based on `dev`. First of the climatisation chain, followed by #853 and then #856.

## Changelog

Climatisation can now be controlled from a climate entity, including the target temperature.

## Test plan

1. Open the device page for a vehicle that reports climatisation.
   - [ ] A `Climatisation` thermostat appears with an off/on mode and a target temperature.
2. With climatisation off, change the target temperature.
   - [ ] The value is stored and the car does not start climatising.
3. Turn it on.
   - [ ] The car starts at the temperature set in step 2.
4. While it runs, check the card against the car.
   - [ ] Heating, cooling and ventilation each show the matching action rather than a generic "on".
5. Restart Home Assistant.
   - [ ] The target temperature is the one you set, not 21.0.
6. On a vehicle that does not report climatisation, reload.
   - [ ] No climate entity is created.

24 tests, and five mutations were verified to fail the guard that owns them, including unwiring the restore from `async_added_to_hass` and starting through the dead legacy action.

Verified against a live vehicle, 4 of 4:

- Setting a temperature while off stored it and did **not** start the car.
- Turning on made the vehicle report `cooling` within about 21 seconds, so the entity shows `HVACAction.COOLING` rather than a generic "on". That value is the car's own status, not an echo of the command.
- The entity followed the car into `heat_cool`.
- Turning off returned it to `off` in about 21 seconds.

One finding from the same session: on API level 0 both start and stop climatisation use the legacy `fs-car` endpoint, which now returns 404. Level 1 uses the Cariad path and works. That is a pre-existing configuration trap rather than anything this PR introduces, but it decides whether the entity can work at all.


lmjabreu:
Do not merge this one on its own. It has a dependency I failed to state when I opened it.

`AudiClimate.async_turn_on` calls `start_climate_control(vin, temp_c=...)` and passes nothing else. On `dev` that makes `has_settings` true, because `temp_c` is not None, and the body is then built with every other field coerced from `None`:

```python
"climatizationAtUnlock": bool(climatisation_at_unlock),   # bool(None) -> False
"windowHeatingEnabled": bool(glass_heating),              # False
"zoneFrontLeftEnabled": bool(seat_fl),                    # False
"zoneFrontRightEnabled": bool(seat_fr),                   # False
```

So turning the climate entity on resets the car's stored window heating, seat zones and climatise-at-unlock. Every time. The settings are what the myAudi app shows and what a departure timer uses, so this is not cosmetic.

The fix is `3e98bbe` in #856, which reads the current settings and writes back only what changed. That is seven PRs downstream of this one.

My test plan here was four live checks that all passed and none of which could have caught it, because not one of them reads the settings back afterwards. That is my error, not the reviewer's.

I would rather move the fix down into this PR, since this is the code that triggers it. Happy to do that, or to hold this until #856 is in, whichever you prefer.

lmjabreu:
Resolved, and this PR is now safe to merge on its own.

The fix moved down into this branch, where the bug is: `build_climatisation_start_body` reads the car's current settings and changes only what the caller actually supplied, so `None` means leave it alone and only an explicit `False` turns something off. The 19-line read helper it needs came with it, since this is the first thing that needs it.

Two things came out of doing that, both worth stating because they contradict what I wrote above.

The fix did not depend on the settings readback in #853, as I assumed. It called `get_climatisation_settings_raw`, which is defined in #856, one further downstream again. So the dependency was worse than I described, not better, and the tests are what found it rather than my reading of it.

And a guard in #853 asserted `int(temp_c)` appears in `start_climate_control`. The read-modify-write moved that arithmetic into the builder, so the guard was inspecting a function that no longer does the truncation and would have passed forever. Retargeted in #853.

Rebased onto current `dev` after #847 landed. 126 tests pass, and `pytest` now runs in CI since #858 merged.

Issue #846, opened 2026-09-04
Could not start device sign-in with Audi. Please try again later.


segran2:
**Describe the bug**
At installation/start-up of the integration after loading from HACS and entering the correct PIN code I get "Could not start device sign-in with Audi. Please try again later."

**To Reproduce**
Bog standard installation of Audi Connect integration.

**Expected behavior**
Expecting a connection Audi Connect web server where I will be asked for username and password.

**Screenshots*

Could not start device sign-in with Audi. Please try again later.

**Logfile**
Can't activate logging as the integration never installs-.

**Your Vehicle Details**
Model: Audi e-tron
Year: 2023
Type Electric
Region EU

<img width="1234" height="514" alt="Image" src="https://github.com/user-attachments/assets/1ed74322-2628-4935-9904-f4bca13400c4" />

Just-inc-ace:
Same here. 

Audi device authorization request failed
Traceback (most recent call last):
  File "/config/custom_components/audiconnect/config_flow.py", line 208, in async_step_device
    response = await self._connection.request_device_code()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/audiconnect/audi_connect_account.py", line 102, in request_device_code
    return await self._audi_service.request_device_code()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/audiconnect/audi_services.py", line 1643, in request_device_code
    raise Exception(
    ...<2 lines>...
    )
Exception: Device authorization request failed: unauthorized_client

minosrex:
I first got a dialogue with my existing installation, asking to re-authorise. However, upon clicking the link, I got:
"{"error":"INVALID_REQUEST","error_description":"Provided request is invalid."}"

I removed and re-added the integration, to get the message as above ("could not start device sign-in").

antoinevalentinHA:
The failure here is `request_device_code()` getting `unauthorized_client` back from VW's `device_authorization` endpoint — the integration sends its `client_id` + the device-code scope, and VW now answers `{"error":"unauthorized_client"}` instead of a `device_code`. That's a client-level refusal of the grant, not a credential or PIN problem, which is why it fails before any sign-in page appears; the config flow just surfaces it as the generic "Could not start device sign-in… try again later."

This looks partial rather than total, so a few things are worth pinning down:

[...]

PR #850, opened 2026-09-05
test: cover the device-code auth state machine

lmjabreu:
Reopens #815, which closed twice without ever being reviewed. Both closures were the base branch being deleted, not a decision about the change: `beta` was removed on 2026-08-05 (reopened a minute later) and again on 2026-08-06, and GitHub auto-closes a PR when its base goes. There are no comments and no reviews on it. `beta` has since been retired in favour of `dev`, so this targets `dev`.

## Overview

The device-code flow from #777 still has no automated coverage, and it keeps needing careful changes around mbb scope and refresh gating. This adds a pytest suite that mocks the `AudiAPI` HTTP surface and drives `AudiService` directly: device-code request and discovery, the RFC 8628 poll outcomes, session finalisation with and without an mbb `refresh_token`, refresh rotation and rejection, and id_token parsing.

The content is unchanged from #815. All 11 tests pass against the current `dev` auth code without modification, which is worth noting given how much moved in between (refresh-token persistence, `AudiAuthError`, `ConfigEntryAuthFailed`).

11 tests, about 1 second, no network.

`pytest.ini` sets `asyncio_mode = auto` and `conftest.py` puts the repo root on `sys.path`. The existing tests on `dev` use `asyncio.run` inside sync tests and are unaffected; the full suite is green.

## Reference

Reopens #815. Covers the flow added in #777.

## Changelog

No user-facing change.

## Test plan

1. `pip install homeassistant pytest pytest-asyncio beautifulsoup4`
2. `pytest`
   - [ ] 11 tests in `tests/test_audi_services_auth.py` pass.
   - [ ] The existing suite still passes alongside them.

To confirm the tests are exercising the real code rather than passing vacuously, change `get_id_token_subject` to prefer `sub` over `email`:
   - [ ] `test_get_id_token_subject_parses_email` fails.

Nothing in the repository runs pytest today, so these only run locally. Happy to add a CI workflow in a follow-up if you would take it, which was also offered on #815.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+asyncio_mode = auto
+testpaths = tests
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,4 @@
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
--- a/tests/test_audi_services_auth.py
+++ b/tests/test_audi_services_auth.py
@@ -1,0 +1,267 @@
+"""Tests for the OAuth Device Authorization Grant auth state machine.
+
+Mocks the full AudiAPI HTTP surface (markets, market config, OIDC discovery,
+device authorization, token endpoint, AZS and mbboauth) and exercises
+AudiService.request_device_code / poll_device_token / login_with_refresh_token /
+get_id_token_subject.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+from typing import Any
+
+import pytest
+
+from custom_components.audiconnect.audi_services import (
+    AudiAuthError,
+    AudiService,
+    DEVICE_CODE_SCOPE,
+)
+
+TOKEN_ENDPOINT = "https://identity.vwgroup.io/oidc/v1/token"
+DEVICE_AUTH_ENDPOINT = "https://identity.vwgroup.io/oidc/v1/device_authorization"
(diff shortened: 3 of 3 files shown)


antoinevalentinHA (approved):
Traced each assertion against the current `dev` auth code and they model it faithfully — the poll error mapping (`expired_token`→`expired`, `access_denied`→`denied`, unmapped→`error`), the id_token `email or sub` fallback, refresh-token rotation, and the AZS→register→auth→vwToken derivation in `_finalize_session`, including the "no mbboauth `refresh_token` → use the auth token as vwToken" branch. The two I care most about are `test_refresh_if_necessary_keys_on_idk_token_without_mbb_refresh` and the rejected-refresh cases — they pin exactly the IDK-independent-of-mbboauth refresh (#782) and the `invalid_grant`→reauth / everything-else→retryable split from #843. Good to have those locked down.

Two notes, neither blocking:

1. **CI doesn't run pytest today** — the checks here are hassfest / HACS / pre-commit only, so this suite (and the nine test files already on `dev`, the #843 classifier included) only protect local runs, nothing at merge time. Your offer to wire a pytest workflow in a follow-up is the piece that makes all of this load-bearing — I'd take it, and I'm happy to review that workflow.

2. **One coverage gap worth a line:** `request_device_code` is only exercised on the success path. Its failure path is what's breaking in the field right now (#846: VW returns `unauthorized_client` with no `device_code`). A case asserting it raises cleanly on that body would lock in the behaviour we're currently leaning on. Fine as a follow-up.

+274/-0, test-only, no production change. LGTM.

PR #854, opened 2026-09-08
feat: gate entities on what the car can do, not on the current poll

lmjabreu:
> Rebased onto `dev` and no longer stacked. The diff is now this change alone: two commits, 10 files. This was previously shown on top of #847, #848, #849, #852 and #853, which is what @antoinevalentinHA had to read past.

## Overview

`is_entity_supported` asks whether a value is present in the poll the integration happens to be holding. That is not the same question as whether the car has the feature, and the difference is not theoretical: a `429 Too Many Requests` on a live vehicle left sixteen entities gone, among them a parking-position sensor for a car whose capability list plainly contains `parkingPosition`. They came back only after a reload.

To be precise about the mechanism, because the first version of this paragraph was not: entity creation happens once, in `async_setup_entry`. A rate-limited poll cannot remove an entity that already exists. What it does is decide the entity set whenever the integration sets up, so the loss happens when a restart or reload lands inside a degraded window, and it persists until the next setup with a good poll.

`userCapabilities` answers the right question and is stable across polls. It was commented out of `JOBS2QUERY` on 2024-03-02 with no explanation in the commit, and has been unused since. This enables and parses it: 61 capabilities on the vehicle it was read from, each with an id and an expiry date, two of them carrying error codes, because a listed capability is not necessarily a working one.

`has_capability` is deliberately three-valued: `True`, `False`, or `None` when the car reports no list. A vehicle that says nothing falls back to exactly the previous behaviour, because reading "did not say" as "not capable" would strip every entity from those cars. That fallback is the half most likely to break people, so it carries the most tests.

Opted in only where the mapping is exact, which is the parking-position family. Any description naming no capability behaves precisely as before, so the blast radius is three entities rather than seventy, and extending it later is one field per description.

One honest limitation: the list is **job-level**, not field-level. It answers "does this car do climatisation", never "does it report `isMirrorHeatingActive`". Entities gated on a single field inside a capability need a different fix and are untouched here.

## Reference

Standalone, based on `dev`. Enables the job disabled in 1c136e0. #855 depends on this PR for the `capability` argument.

## Changelog

Entities for features the car supports are no longer removed when a poll arrives without their data.

## Test plan

1. On a vehicle that reports a capability list, open the device page.
   - [ ] The parking position sensors exist.
2. Force a partial poll (rate limit, or a vehicle that has not reported position recently), then restart.
   - [ ] `park_time` reads `unknown` rather than disappearing.
   - [ ] Entities whose descriptions name no capability behave as they did before.
3. On a vehicle that reports no capability list.
   - [ ] Nothing changes: entities appear exactly as they did previously.

Verified on a live vehicle. Before this change, a poll without position data left `sensor.park_time` unavailable and the entity uncreated; after it, the sensor exists and reads `unknown` while `device_tracker.position` and `position_last_updated` report normally. The unavailable count on that vehicle went from sixteen to four, and all four remaining are intended: two engine buttons and a window heating switch removed earlier in the stack, plus one sensor gated on a field rather than a capability, which this change deliberately does not address.

18 tests. Three mutations verified: re-commenting the job, ignoring the capability, and collapsing the three-valued check to a boolean. One test guards the test suite itself, pinning the stub's `has_capability` against the real one after the first version of that stub disagreed with the implementation about what an empty list means.


Kolbi:
@lmjabreu I am currently in vacation & will have a look into all PRs after it.

antoinevalentinHA (commented):
Read this as commit `1aab64a` (the capability gate itself), not the stacked diff. The design is right and it fixes something real: gating entity creation on `userCapabilities` rather than whatever the current poll happens to hold. I've had the parking-position sensor disappear after a partial poll and only come back on reload — this is exactly that. The three-valued `has_capability` (True / False / None = "said nothing") is the correct shape, and the empty-list-isn't-"incapable" fallback is the half that protects cars that never report the list. The test file is thorough, and `test_the_real_has_capability_matches_this_stub` guarding the stub against drift is a nice touch.

One correctness question before I'd approve:

`_parseUserCapabilities` appends its state with `measure_time: None` — the first state in the codebase to do so; both existing append paths (`_appendState`, `_tryAppendStateWithTs`) guard against a falsy ts. It then flows into the `last_update_time` scan in `audi_connect_account.py`, which does `parse_datetime(state.get("measure_time"))`. That loop guards the *result* (`if new_time:`) but not the call, and `parse_datetime(None)` raises `TypeError` in current HA rather than returning None. Since you ran this on a live car, either your runtime tolerates it or this state reaches a different path than I traced — can you confirm? If it does hit that loop, skipping a None ts there (or leaving this statement-about-the-vehicle out of the freshness scan) would settle it, and a small test around that loop would be worth having.

Minor, and about the stack rather than this change: the commit labelled `style: apply pre-commit fixes` (`10e173c`) actually removes the charger switch and the timed-charging button and adds the charge-mode comment — substantive, and belonging to #847/#852. Worth rewording so #854's own history doesn't carry user-facing removals under a "style" label.

Logic and tests are otherwise solid — happy to approve once the `measure_time` path is confirmed.

lmjabreu:
> `parse_datetime(None)` raises `TypeError` in current HA rather than returning None. Since you ran this on a live car, either your runtime tolerates it or this state reaches a different path than I traced — can you confirm?

Confirmed, `parse_datetime` resolves to `.util`, not `homeassistant.util.dt`:

```python
from .util import get_attr, log_exception, parse_datetime, parse_float, parse_int
```

It returns `None` for anything that isn't a `datetime` or a `str`, so the `if new_time:` guard covers it. HA's does raise, so the trace was right up to the import.

Ran it over the Q8 e-tron fixture with a control, driving the real `update_vehicle_statusreport`:

```
MUST-FIRE: real payload (has userCapabilities)
  states with measure_time=None : ['userCapabilities', 'userCapabilitiesImpaired']
  vehicle.state keys assigned   : 42
  ERROR logs                    : 0

CONTROL: same payload, userCapabilities removed
  states with measure_time=None : []
  vehicle.state keys assigned   : 40
  ERROR logs                    : 0
```

42 against 40 is the two capability states, and the assignment loop below the scan completes both times, so nothing is being swallowed by the broad `except Exception`.

Hardened it anyway in `31b0e0a`, since it shouldn't rest on which module the name resolves to. The loop now skips a falsy ts, and there's a test around it, which there wasn't before. Writing that turned up something neither of us had: the `data_fields` loop above carries `datetime` objects rather than strings, so a wholesale swap to HA's `parse_datetime` would break there too. The test deliberately doesn't cover that, it pins the one property you raised (a scan that survives `None`) rather than asserting a promise about field types that the integration never made.

Good catch on the commit message. `10e173c` carries the charger switch and timed-charging removals, which belong to #847/#852. I'll reword it on the next restack rather than now, so `1aab64a` stays the SHA you reviewed.

antoinevalentinHA (approved):
That's the answer — thanks for tracing it to `.util.parse_datetime` and for the fixture run. Confirmed on my side: the import at `audi_connect_account.py:16` is the local `.util` one, which returns `None` for a non-str/datetime, so the `if new_time:` guard holds today. The reason it was worth raising is exactly what `31b0e0a` captures — the scan survives by which module the name happens to resolve to, and `test_scan_survives_a_strict_parse_datetime` pins that so a later swap to HA's `parse_datetime` can't silently stop every state update. Skipping a falsy ts is the right minimal fix, and the `data_fields`-carry-`datetime`s note is a good one to have on record.

Reword-on-next-restack is fine — keeping `1aab64a` as the reviewed SHA is the right call. Approving; merge waits on the stack base (#847–#853).

antoinevalentinHA:
Thanks for de-stacking it — reading it as the change alone is much better, and it's the same shape I reviewed (the charger/timed-charging removal correctly dropped out with the rebase, since that belonged to #847/#852). Approval stands.

One practical note: now that #847 has landed on `dev`, GitHub flags this as conflicting — a rebase onto current `dev` should clear it, and then it's good to go from my side.

lmjabreu:
> now that #847 has landed on `dev`, GitHub flags this as conflicting

Done. Rebased onto current `dev` at `53509ec`, mergeable again and green. The only conflict was the readme, where #847's new Controls section and this one's sit next to each other; both kept.

Worth knowing: #858 landed too, so `pytest` now runs on pull requests. This is the first PR to go through it, 113 tests in 50s. Before that the checks were hassfest, HACS and the semantic-title lint, so a green tick said nothing about the code.

PR #856, opened 2026-09-08
feat: make the climatisation settings writable

lmjabreu:
> Rebased onto current `dev`, on top of #849 and #853, so the diff shows their commits until they merge; the commits to review here are `feat: make the climatisation settings writable` and `fix: a settings switch must not snap back while the car catches up`.
>
> **Merge order.** Three independent lines, so nothing here needs coordinating across them:
>
> - #855, on its own
> - #848 → #852
> - #849 → #853 → #856, then #857 (draft) last
>
> Every one of these edits `readme.md`, so each merge makes the others conflict. That is expected rather than a problem with the PR. I rebase and push as soon as I see it, normally the same working day, so a conflicting state here is stale rather than stuck.

## Overview

#853 recorded that no settings endpoint existed, because a `GET` to `climatisation/settings` returned 404. That inference was wrong, and the honk-and-flash work in #855 demonstrated why on the same API: this BFF answers 404 rather than 405 for a path with no `GET` route, and `honkandflash` 404s on `GET` while being perfectly real for `POST`.

There is an endpoint:

```
PUT {bff}/vehicle/v1/vehicles/{vin}/climatisation/settings
```

carrying `targetTemperature`, `climatizationAtUnlock`, `windowHeatingEnabled`, `heaterSource` and the four `zone*Enabled` flags, which is exactly the block #853 parses. Contract from [WeConnect-python](https://github.com/tillsteinbach/WeConnect-python), which drives the same BFF.

**Read-modify-write**, because the endpoint replaces the object rather than merging into it. Toggling a heated seat has to round-trip the target temperature and the heater source or it clears them. Same hazard as the charging profile write in #839, same answer, and `build_climatisation_settings_update` is a pure module-level core so field preservation is testable without the HTTP boundary. `carCapturedTimestamp` is stripped, since it is the server saying when it last observed the settings rather than a setting.

A field outside the known set is refused rather than sent, so a typo becomes an error rather than a field the car quietly ignores.

The six diagnostic binary sensors #853 added for these fields become switches, which carry the same readback plus control. The pre-existing `glass_surface_heating` sensor stays: it reports whether mirror heating is running, which is not the same as whether it is enabled.

Worth noting this offers more than the myAudi app does. There, a heated seat can only be enabled while starting the air conditioning; the API has no such restriction.

## Reference

Eighth of a stack, based on #855. Corrects the endpoint claim in #853. Contract from WeConnect-python.

## Changelog

Climatisation settings such as seat zones and window heating can now be changed from switches, without starting climatisation.

## Test plan

1. Open the device page for a vehicle that reports climatisation settings.
   - [ ] Switches appear for window heating, climatisation at unlock, and the seat zones the car has.
2. Note the climate entity's target temperature, then toggle a seat zone.
   - [ ] The zone changes.
   - [ ] The target temperature is unchanged.
   - [ ] The other zones are unchanged.
3. On a vehicle reporting only front zones.
   - [ ] No rear zone switches are created.

Verified on a live vehicle. With the target temperature at 19.0 and every zone off, turning the front left zone on left the temperature at 19.0 and the other switches untouched, then turning it off restored the original state. That is the read-modify-write doing its job: a lone-field PUT would have cleared the temperature.

16 tests. Five mutations verified, including sending only the changed field, PUTting without reading first, and closing the switch lambdas over the loop variable so that every switch would write the same field.


Kolbi:
Please resolve conflicts

Kolbi:
@lmjabreu can you please double check if my conflict resolving is alright? Thanks in advance.

lmjabreu:
> @lmjabreu can you please double check if my conflict resolving is alright?

Looks right apart from two hunks the merge lost. Put them back in da3c39e: the removal of the six diagnostic binary sensors the switches replace (otherwise each setting shows twice on the device page), and the login guard at the top of `set_climatisation_setting` that the other methods have. Everything else matched, tests and ruff pass. Thanks for sorting the rest 🙏

PR #857, opened 2026-09-13
fix: keep an entity a thin poll cannot see

lmjabreu:
> Rebased onto current `dev` at `5ca590f`, cleanly. Everything this depended on has merged (#847, #848, #849, #852, #853, #855, #856), so the diff is now this change alone: two commits, 10 files.
>
> The three defects an independent review found are fixed in the second commit. See the thread below.

## Overview

#854 stopped a rate-limited poll deleting entities by gating creation on the car's capability list. That list is job-level: it answers "does this car do climatisation", never "does it report `isMirrorHeatingActive`". So two gaps survived it. An entity backed by a single field, inside a capability the car has, is still dropped when a poll arrives without that field. And a car that reports no capability list at all is protected by nothing, because the gate has to fall back to the old presence check there or it would strip every entity from those vehicles.

The registry gets a vote. An entity already registered for this car is one this integration created on an earlier, better poll. Recreating it costs nothing and it reports unavailable while the value is missing, which is the truthful state. Losing it takes the user's history, their dashboard card and any automation that names it.

Order of decision in `should_create_entity`: the car says no, stop. The car says yes, create. The value is in this poll, create. It was registered before, create. Otherwise no.

The car's own no deliberately outranks the registry. A capability the car does not list is a statement about the vehicle rather than about this poll, so a stale entry cannot resurrect a control for a feature the car does not have.

Eight of the ten platforms now build their unique_id through one `entity_unique_id` helper (`button.py` and the API-rate-limit sensor keep theirs, and neither is in the rescue). The lookup only finds an entity if it asks for the exact string that entity was created with, so that format cannot be spelled out in nine places.

The capability check only bites where a description names a capability, which today is the parking-position family. Everywhere else the rescue rests on the registry alone. That was true of the first version too and went unsaid.

Home Assistant core pairs this pattern with a coordinator listener that adds entities when their data first arrives (`withings/sensor.py`). This carries only the registry half, so a fresh install during a degraded window still needs one manual reload.

Two gaps are left deliberately. `button.py` is untouched: its gates are the S-PIN and the car type, which are configuration and vehicle facts rather than poll contents, so a rescue there would be wrong, because removing the S-PIN should remove the buttons. And the per-profile charge-target sliders are gated on the profile list being present rather than on this function, so they carry the same hazard; recreating one needs its name as well as its id, which is its own change.

## Reference

Ninth of a stack, based on #856. Extends the capability gate in #854 to the field level.

## Changelog

An entity is no longer removed because a single poll arrived without its value.

## Test plan

1. On a vehicle with entities already created, cause a partial poll (a rate limit, or a job answering with an error) and reload the integration inside that window.
   - [ ] Entities fed by the missing job are still present.
   - [ ] They report unavailable rather than disappearing.
2. Restore a healthy poll.
   - [ ] Those entities report values again, with their history intact.
3. On a fresh install of a car that lacks a feature.
   - [ ] No entity is created for it.

The risk here is the opposite of the bug: inventing entities that should not exist. 16 tests cover both directions, and each rescue case has its control. The must-fire case is a field absent from the poll with the entity registered; its control is the identical call against an empty registry. A second control registers the entity under a different VIN, so a two-car account cannot leak entities between vehicles. The car's explicit no is tested against a registry that does hold the entity.

The unique_id format is pinned across all nine platform and key shapes, because a change there would orphan every existing entity on every install, silently, while the new ones looked fine.

Two mutations verified with the diffstat printed beside each run: removing the registry vote fails the must-fire test, and changing the unique_id format fails nine. 233 tests pass, ruff clean.


lmjabreu:
Holding this as a draft until the rest of the set merges, and it is worth saying why rather than leaving it sitting there.

This PR edits every platform's `async_setup_entry`, so it touches `select.py` (#852), `climate.py` (#849), and the switches from #853 and #856. Rebasing it onto any one of those chains conflicts in five files, and `select.py` does not exist on `dev` at all yet. It is the one change in the set that genuinely couples the others, which is exactly what an independent reviewer said about it before I had worked it out.

So the order is: #848, #852, #849, #853, #856, then one rebase of this onto `dev` and it is clean.

Three things to fix here before it is ready for review, all found by that same review and all confirmed against the code:

- **The lock becomes immortal.** `lock_supported` is `doors_trunk_status_supported and self._audi_service._spin is not None`. I excluded `button.py` on the grounds that removing the S-PIN should remove those controls, then routed the lock, which has the same S-PIN gate, straight through the registry rescue. A rescued lock also reports Unlocked, since `is_locked` compares against `"Locked"` and missing data is not that.
- **"Reports unavailable" is not true.** The body says a rescued entity reads unavailable, which would be the truthful state. No platform overrides `available`, and `CoordinatorEntity.available` is `last_update_success`, which is True on a thin but successful poll. So a rescued switch reads OFF and a rescued sensor reads unknown. A config switch showing OFF for a setting that may be ON is worse than the entity being absent.
- **The capability check is inert on six of eight platforms.** Only `sensor.py` descriptions carry a `capability`, so "the car's own no outranks the registry" only fires for sensors and the device tracker. Everywhere else the rescue is unconditional.

Also worth doing: Home Assistant core pairs this pattern with a coordinator listener that adds entities when their data first arrives (`withings/sensor.py`). This carries only the registry half, so a fresh install during a degraded window still needs a manual reload.

lmjabreu:
Unblocked. Everything this depended on is now in `dev`, so the hold above no longer applies.

Rebased onto `dev` at `5ca590f`, cleanly, no conflicts. It is two commits and 10 files now rather than the 36 it showed while sitting on the old stack base. 259 tests pass, ruff and format clean.

The three defects named above are fixed in the second commit:

- **The lock is no longer immortal.** `should_create_entity` takes a `configured` argument and `lock.py` passes whether the S-PIN is set, so removing it removes the lock, which is the same reasoning that keeps `button.py` out of the rescue.
- **"Reports unavailable" is now true**, where before it was an assertion with nothing behind it. `AudiEntity.available` returns False when the backing attribute is absent.
- **The capability limit is stated** rather than implied: it only bites where a description names a capability, which today is the parking-position family.

Worth noting what #856 merging changed about this PR's value. The six climatisation settings switches it added are `EntityCategory.CONFIG`, backed by a single field each, and their `is_on` is `bool(value)`. That is exactly the case the availability fix exists for. Checked against the shipped descriptions on `dev`:

```
climatisation_window_heating: value absent -> available=False, is_on=False
climatisation_window_heating: value False  -> available=True,  is_on=False
```

Without it, a settings switch whose value a poll does not carry reads "off" for a setting that may well be on, and an automation acts on that. With it, it reads unavailable, which is what Home Assistant has the word for.

Still a draft, pending a look from someone before I mark it ready.

PR #864, opened 2026-09-18
fix: a 401 from the legacy preheater endpoint is not an error

### LINKED ISSUE #862: preheater error in logging
I get a preheater error in the logging, but the integration doesn't show any sensors or diagnostic options.

Steps to reproduce the behavior: ?

**Expected behavior**
If there was a sensor, I could disable it, if it produced an error and/or didn't work

**Logfile**
This error originated from a custom integration.

Logger: custom_components.audiconnect.audi_connect_account
Source: custom_components/audiconnect/audi_connect_account.py:1057
Integration: Audi Connect (documentation, issues)
First occurred: 21:33:57 (1 occurrence)
Last logged: 21:33:57

Unable to obtain the vehicle preheater state for WAxxxx: 401, message='Unauthorized', url='https://fal-3a.prd.eu.dp.vwg-connect.com/fs-car/bs/rs/v1/Audi/DE/vehicles/WAxxxx/status'
  File "/config/custom_components/audiconnect/audi_connect_account.py", line 1057, in update_vehicle_preheater
    result = await self._audi_service.get_preheater(self._vehicle.vin)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/audiconnect/audi_services.py", line 125, in get_preheater
    return await self._api.get(
           ^^^^^^^^^^^^^^^^^^^^
        f"{await self._get_home_region(vin.upper())}/fs-car/bs/rs/v1/{self._type}/{self._country}/vehicles/{vin.upper()}/status"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/config/custom_components/audiconnect/audi_api.py", line 188, in get
    return await self.request(
           ^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/config/custom_components/audiconnect/audi_api.py", line 156, in request
    raise ClientResponseError(
    ...<4 lines>...
    )
aiohttp.client_exceptions.ClientResponseError: 401, message='Unauthorized', url='https://fal-3a.prd.eu.dp.vwg-connect.com/fs-car/bs/rs/v1/Audi/DE/vehicles/WAxxxx/status'

**Your Vehicle Details**
Model: e-tron 55
Year: 2021
Type Electric
Region EU

lmjabreu:
Fixes #862.

## Overview

The legacy `fs-car` status endpoint answers 401 for some accounts, and every restart logged a full traceback with nothing the user could switch off. The handler already treats 403 and 502 as transient (debug log, feature stays on, next poll retries); 401 was falling through to the generic error path. Now it joins the transient branch. It is not disabled like a 404, because a 401 can also be a temporary API ban and disabling until restart would be the wrong answer to that.

The message also carried the **full VIN**, as did six others of the same shape across the status-report, preheater and charger paths; the reporter had redacted it by hand before pasting. A VIN identifies a car and its owner. All seven now use the last four characters, through one `_redacted_vin` property in place of five identical local expressions. The debug lines in the same functions already did this; the error lines did not.

No entity to disable the poll, which the issue asked for: once the 401 is quiet there is nothing left to disable.

## Reference

Fixes #862. Independent of #863.

## Changelog

A preheater endpoint the account cannot reach no longer logs an error on every restart, and error messages no longer include the full VIN.

## Test plan

1. On an account whose legacy preheater endpoint answers 401, restart Home Assistant.
   - [ ] No `Unable to obtain the vehicle preheater state` error in the log.
   - [ ] At debug level, one `PREHEATER: Received status 401` line per poll, and the poll keeps running.
2. Force a genuine failure (a 500, or the host unreachable).
   - [ ] The error is still logged once.
   - [ ] The VIN in it shows only its last four characters.

8 tests: the 401 is quiet and keeps the feature on, it is retried on the next poll, a 404 still disables and is not retried, a 500 still logging and keeping the feature as the control, and the three error paths asserted never to carry the VIN.


Kolbi:
I am just thinking if 401 really means not usable or maybe just temporarly account banned for using APIs?

lmjabreu:
> I am just thinking if 401 really means not usable or maybe just temporarly account banned for using APIs?

Ah yes, it was just one report so we don't actually know if it's permanent, makes sense. Changed in 9fb7108 to match the 403/502 handling: debug log, poll stays on, so it recovers if the ban lifts. The traceback from #862 goes away either way.

